PR #1, opened 2018-07-09
Initial architecture setup

raigex:


agrosner:
MR mostly looks great. just ensure you're using latest of dependencies in project, remove stub unit tests and android tests, ensure you don't have extraneous resources and itll look good to merge.

PR #7, opened 2018-07-17
Splash

raigex:


db/src/main/kotlin/edu/artic/db/AppConverters.kt
```diff
@@ -0,0 +1,145 @@
...
+        return tourCategoryListAdapter.toJson(objects)
+    }
+
+    @TypeConverter
+    fun stringToTourCategoryList(json: String?): List<ArticTourCategory> {
+        return safeList(json, tourCategoryListAdapter)
+    }
+
+    private inline fun <T> safeList(data: String?, adapterGetter: JsonAdapter<List<T>>): List<T> {
+        return data?.let { adapterGetter.fromJson("") } ?: emptyList()
```
agrosner: you're passing `fromJson("')`. `data` is not parsed!

db/src/main/kotlin/edu/artic/db/AppDataApi.kt
```diff
@@ -1,13 +1,13 @@
...
+import edu.artic.db.models.ArticAppData
+import edu.artic.db.models.ArticEvent
+import edu.artic.db.models.ArticExhibition
 import edu.artic.db.progress.Constants
 import io.reactivex.Observable
 import retrofit2.Response
-import retrofit2.http.GET
-import retrofit2.http.HEAD
-import retrofit2.http.Header
+import retrofit2.http.*
```
agrosner: no `*` import.

db/src/main/kotlin/edu/artic/db/AppDataApi.kt
```diff
@@ -16,5 +16,21 @@ interface AppDataApi {
 
 
     @GET("appData-v2.json")
-    fun getBlob(@Header(Constants.DOWNLOAD_IDENTIFIER_HEADER) header: String): Observable<Result<ArticBlobData>>
+    fun getBlob(
+            @Header(Constants.DOWNLOAD_IDENTIFIER_HEADER) header: String
```
agrosner: `Constants` seems a little generic and god-object like. Make it more specific like `ApiConstants`

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
-    fun getBlob(): Observable<AppDataState> {
+        //First load app data. if empty is returned, do nothing, otherwise download everything else.
+        return Observable.create<ProgressDataState> { observer ->
+            getBlob().subscribe({
+                when (it) {
+
+                    is ProgressDataState.Done<*> -> {
+                        loadPostAppData()
+                                .subscribe({ amountDownload ->
+                                    Log.d("AppDataManager", "Amount downloaded $amountDownload")
```
agrosner: this exact set of code is duplicated below. Make a method consolidating it.

agrosner: it's same as `Empty`.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                    }
+                    is ProgressDataState.Downloading -> {
+                        observer.onNext(ProgressDataState.Downloading(it.progress * .33f))
+                    }
+                }
+
+            }, {
+                observer.onError(it)
+            }, {
+
```
agrosner: if this is `onComplete`, maybe its better to use named parameters on this method to make it clear what its doing.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                        //Save last downloaded headers
+                        appDataState.headers[HEADER_LAST_MODIFIED]?.let {
+                            appDataPreferencesManager.lastModified = it[0]
+                        }
+
+                        val result = appDataState.result as ArticAppData
+
+                        try {
+                            appDatabase.dashboardDao.setDashBoard(result.dashboard)
+                        } catch (e: Throwable) {
```
agrosner: try catching `throwable` suppresses any  exception thrown here. Should we catch specific exceptions here? or throw exception on Debug builds so we can catch errors here fast and early. I can foresee us seeing blank data and not realizing it until much later and spending a lot of time figuring out why something is going wrong.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                            e.printStackTrace()
+                        }
+                        try {
+                            appDatabase.generalInfoDao.setGeneralInfo(result.generalInfo)
+                        } catch (e: Throwable) {
+                            e.printStackTrace()
```
agrosner: same as above.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                        }
+                        val galleries = result.galleries
+                        if (galleries?.isNotEmpty() == true) {
+                            try {
+                                appDatabase.galleryDao.addGalleries(galleries.values.toList())
+                            } catch (e: Throwable) {
+                                e.printStackTrace()
```
agrosner: same as above.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                            }
+                        }
+                        val objects = result.objects
+                        if (objects?.isNotEmpty() == true) {
+                            try {
+                                appDatabase.objectDao.addObjects(objects.values.toList())
```
agrosner: same as above.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                            } catch (e: Throwable) {
+                                e.printStackTrace()
+                            }
+                        }
+                        val audioFiles = result.audioFiles
+                        if (audioFiles?.isNotEmpty() == true) {
+                            try {
+                                appDatabase.audioFileDao.addAudioFiles(audioFiles.values.toList())
+                            } catch (e: Throwable) {
+                                e.printStackTrace()
```
agrosner: same as above.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                            e.printStackTrace()
+                        }
+                    }
+                    return@flatMap appDataState.asObservable()
+                }
+    }
+
+    private fun loadPostAppData(): Observable<Int> {
+        return Observable.create { observer ->
+            val maxDownloads = 2
```
agrosner: what are we doing here? what is max downloads.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+            var currentDownloads = 0
+            observer.onNext(currentDownloads)
+            getExhibitions().subscribe({
```
agrosner: if you're paralleling work, couldnt we use `zip()`?

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
-                            .appendPattern("E, dd MMM yyyy HH:mm:ss O")
-                            .toFormatter()
-                    if (it.containsKey("last-modified")) {
-                        val lastModified = it["last-modified"]?.get(0)!!
-                        val newLocalDateTime = LocalDateTime.parse(lastModified, dtf)
-                        if(newLocalDateTime.isAfter(/*TODO: get from storage*/ LocalDateTime.parse("Thu, 12 Jul 2018 16:56:12 GMT", dtf))) {
-                            return@flatMap serviceProvider.getBlob()
-                        } else {
-                            return@flatMap Observable.just(AppDataState.Empty)
+                    when (it) {
```
agrosner: what is `it` here? `response` might make it clearer. not sure.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                        is ProgressDataState.Done<*> -> {
```
agrosner: maybe a generic extension filter method like `flatMapDone()` with the proper type casting would work here as it could cleanup code and reduce boilerplate.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -1,38 +1,221 @@
...
+                    it.asObservable()
+                }
+    }
+
+
+    private fun getEvents(): Observable<ProgressDataState> {
+        return serviceProvider.getEvents()
+                .flatMap {
+                    when (it) {
+                        is ProgressDataState.Done<*> -> {
```
agrosner: same as above. it would cleanup duplicated code. maybe in a `retrofit-rx-extensions` module or something.

db/src/main/kotlin/edu/artic/db/ArticResult.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.squareup.moshi.JsonClass
+
+
+@JsonClass(generateAdapter = true)
+data class ArticResult<T>(
+        @Json(name = "pagination") val pagination : ResultPagination,
+        @Json(name = "data") val data : List<T>
+)
+
+data class ResultPagination(
```
agrosner: you're using reflection here possibly? add `JsonClass(generateAdapter = true)`

db/src/main/kotlin/edu/artic/db/MoshiAdapters.kt
```diff
@@ -16,6 +20,30 @@ inline fun getMoshi(configureBlock: (Moshi.Builder.() -> Unit) = {}): Moshi =
 internal val apiErrorMoshi: Moshi = getMoshi()
 
 fun Moshi.Builder.registerAdapters() = apply {
-    add(ApiJsonAdapterFactory.INSTANCE)
+//    add(ApiJsonAdapterFactory.INSTANCE)
```
agrosner: commented out code.

db/src/main/kotlin/edu/artic/db/RetrofitAppDataServiceProvider.kt
```diff
@@ -40,4 +73,171 @@ class RetrofitAppDataServiceProvider(@Named(ApiModule.RETROFIT_BLOB_API) retrofi
 
         }
     }
+
+    override fun getExhibitions(): Observable<ProgressDataState> {
+        return Observable.create { observer ->
+            var url = dataObject.dataApiUrl + dataObject.exhibitionsEndpoint
+            Log.d("AppDataServiceProvider", "url: $url")
```
agrosner: use timber.

db/src/main/kotlin/edu/artic/db/RetrofitAppDataServiceProvider.kt
```diff
@@ -40,4 +73,171 @@ class RetrofitAppDataServiceProvider(@Named(ApiModule.RETROFIT_BLOB_API) retrofi
...
+                    "title",
+                    "short_description",
+                    "legacy_image_mobile_url",
+                    "legacy_image_desktop_url",
+                    "gallery_id",
+                    "web_url",
+                    "aic_start_at",
+                    "aic_end_at"
+            )
+            postParams["sort"] = listOf("aic_start_at", "aic_end_at")
```
agrosner: why arent we using `retrofit` here??

db/src/main/kotlin/edu/artic/db/RetrofitAppDataServiceProvider.kt
```diff
@@ -40,4 +73,171 @@ class RetrofitAppDataServiceProvider(@Named(ApiModule.RETROFIT_BLOB_API) retrofi
...
+        return Observable.create { observer ->
+            var url = dataObject.dataApiUrl + dataObject.eventsEndpoint
+            Log.d("AppDataServiceProvider", "url: $url")
+            if (!url.contains("/search")) {
+                url += "/search"
+            }
+            url += "?limit=500"
+
+            val postParams = mutableMapOf<String, Any>()
+            postParams["fields"] = listOf(
```
agrosner: same q as above.

splash/src/main/kotlin/edu/artic/splash/SplashActivity.kt
```diff
@@ -13,16 +17,33 @@ class SplashActivity : BaseViewModelActivity<SplashViewModel>() {
...
+                        is Navigate.Back -> {
+
+                        }
+                    }
+                }.disposedBy(disposeBag)
+    }
+
+    override fun onPause() {
+        super.onPause()
+        disposeBag.clear()
```
agrosner: maybe we have a separate `navDisposeBag` that we use instead for bindings that occur only when screen is visible? in that case we probably want it in `onStop()/onStart()` so in multiwindow.

splash/src/main/kotlin/edu/artic/splash/SplashViewModel.kt
```diff
@@ -1,31 +1,48 @@
...
+import edu.artic.db.ProgressDataState
+import edu.artic.viewmodel.NavViewViewModel
+import edu.artic.viewmodel.Navigate
+import io.reactivex.subjects.PublishSubject
 import javax.inject.Inject
 
-class SplashViewModel @Inject constructor(appDataManager : AppDataManager) : BaseViewModel() {
+class SplashViewModel @Inject constructor(appDataManager: AppDataManager) : NavViewViewModel<SplashViewModel.NavigationEndpoint>() {
+
+    sealed class NavigationEndpoint {
```
agrosner: I dont think this goes here? it should belong somewhere more global like a `nav-actions` module or something that other classes if need be can depend on that directly.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
raigex: already changed, guess I didn't push it up. thanks

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
raigex: So the try catch stuff was not passing along data to onError's before I can remove it for this build.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
raigex: So it's not just the blob we have to download.
We also have to download the exhibits (different then blob exhibits) and events json

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
raigex: yea that is better thanks. replaced.

db/src/main/kotlin/edu/artic/db/MoshiAdapters.kt
raigex: was removed just not pushed up

splash/src/main/kotlin/edu/artic/splash/SplashViewModel.kt
raigex: so NavigationEndpoint is something very specific to a ViewModel in our current implementation of the navigation. so each view model will only have stuff like this for now. We can change it later, but mostly because we don't get how your router stuff works right now.

splash/src/main/kotlin/edu/artic/splash/SplashActivity.kt
raigex: hmm i guess we can do that.

PR #10, opened 2018-07-18
[Exhibitions] - Reconciles exhibitions to cms exhibitions

sam33rdhakal:
Reconciles the exhibitions using JSON blob's exhibitions.

### DIFF
--- a/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
+++ b/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
@@ -99,6 +99,7 @@ class AppDataManager @Inject constructor(
 
                         val exhibitionsCMS = result.exhibitions
                         if (exhibitionsCMS?.isNotEmpty() == true) {
+                            appDatabase.exhibitionCMSDao.clear()
                             appDatabase.exhibitionCMSDao.addCMSExhibitions(exhibitionsCMS)
                         }
 
@@ -144,16 +145,11 @@ class AppDataManager @Inject constructor(
                                  */
                                 appDatabase.exhibitionCMSDao
                                         .getAllCMSExhibitions()
-                                        .subscribe {
+                                        .subscribe { cmsExhibitionList ->
                                             @Suppress("UNCHECKED_CAST")
                                             val list = (result as ArticResult<ArticExhibition>).data
-                                            val mapExhibitionByID = HashMap<String, ArticExhibition>()
-
-                                            list.forEach {
-                                                mapExhibitionByID[it.id.toString()] = it
-                                            }
-
-                                            it.forEach {
+                                            val mapExhibitionByID = list.associateBy { it.id.toString() }
+                                            cmsExhibitionList.forEach {
                                                 mapExhibitionByID[it.id]?.order = it.sort
                                             }
                                             appDatabase.exhibitionDao.updateExhibitions(list)
--- a/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
+++ b/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
@@ -93,7 +93,13 @@ class AppDataManager @Inject constructor(
 
                         val tours = result.tours
                         if (tours?.isNotEmpty() == true) {
+                            appDatabase.tourDao.clear()
                             appDatabase.tourDao.addTours(tours)
+                        }
+
+                        val exhibitionsCMS = result.exhibitions
+                        if (exhibitionsCMS?.isNotEmpty() == true) {
+                            appDatabase.exhibitionCMSDao.addCMSExhibitions(exhibitionsCMS)
                         }
 
                         val mapAnnotations = result.mapAnnotations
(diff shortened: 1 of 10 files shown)


raigex:
Looks good to me @agrosner what do you think?

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -93,9 +93,15 @@ class AppDataManager @Inject constructor(
 
                         val tours = result.tours
                         if (tours?.isNotEmpty() == true) {
+                            appDatabase.tourDao.clear()
                             appDatabase.tourDao.addTours(tours)
                         }
 
+                        val exhibitionsCMS = result.exhibitions
+                        if (exhibitionsCMS?.isNotEmpty() == true) {
+                            appDatabase.exhibitionCMSDao.addCMSExhibitions(exhibitionsCMS)
```
agrosner: Do we wanna clear this out before adding? I think soo.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -133,9 +139,25 @@ class AppDataManager @Inject constructor(
...
-                                appDatabase.exhibitionDao.updateExhibitions((result as ArticResult<ArticExhibition>).data)
-
+                                /**
+                                 * Update the sort order of the exhibitions according to the ArticExhibitionCMS
+                                 */
+                                appDatabase.exhibitionCMSDao
+                                        .getAllCMSExhibitions()
+                                        .subscribe {
+                                            @Suppress("UNCHECKED_CAST")
+                                            val list = (result as ArticResult<ArticExhibition>).data
```
agrosner: this cast is strange. why not cast it further up to the top where its used?

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -133,9 +139,25 @@ class AppDataManager @Inject constructor(
...
+                                            val mapExhibitionByID = HashMap<String, ArticExhibition>()
```
agrosner: don't directly use `HashMap` constructor. use the `mutableMapOf()`

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -133,9 +139,25 @@ class AppDataManager @Inject constructor(
...
+
+                                            list.forEach {
```
agrosner: we can use kotlin collection operators here. just do `list.associateBy { it.id.toString() } `

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -133,9 +139,25 @@ class AppDataManager @Inject constructor(
...
+                                                mapExhibitionByID[it.id.toString()] = it
+                                            }
+
+                                            it.forEach {
+                                                mapExhibitionByID[it.id]?.order = it.sort
```
agrosner: no nested `it`. confusing here. use direct name on the outer object.

db/src/main/kotlin/edu/artic/db/daos/ArticExhibitionCMSDao.kt
```diff
@@ -0,0 +1,21 @@
...
+import edu.artic.db.models.ArticExhibitionCMS
+import io.reactivex.Flowable
+
+@Dao
+interface ArticExhibitionCMSDao {
+
+    @Query("DELETE FROM ArticExhibitionCMS")
+    fun clear()
+
+    @Query("select * from ArticExhibitionCMS")
```
agrosner: one is caps other is lowercase. choose a SQLite syntax pattern and stick with it.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
raigex: We can cast it there ... but at this time we don't really need to use it anywhere higher, so we cast it when necessary

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
sam33rdhakal: Fixed.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
sam33rdhakal: Updated.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
sam33rdhakal: Done.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
sam33rdhakal: Refactored.

db/src/main/kotlin/edu/artic/db/daos/ArticExhibitionCMSDao.kt
sam33rdhakal: Updated all of them to lower case.

raigex:
@agrosner changes were made except for 1 (the casting) will merge in and move the casting higher if we use that code at an earlier point

PR #15, opened 2018-07-20
Add View All Events flow

raigex:


events/src/main/kotlin/edu/artic/events/AllEventsFragment.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlin.reflect.KClass
+
+class AllEventsFragment : BaseViewModelFragment<AllEventsViewModel>() {
+
+    override val viewModelClass: KClass<AllEventsViewModel>
+        get() = AllEventsViewModel::class
+    override val layoutResId: Int
+        get() = R.layout.fragment_all_events
+    override val title: String
+        get() = "Events" // TODO: add to strings or figure out language stuff
```
sam33rdhakal: I have added all the strings requiring localizations to string resource (base module) file so that all the strings are at same place.

PR #35, opened 2018-08-08
Map Refactor

raigex:


raigex:
The only thing left to do with this that can be done at a later time, is fix up the flow so that the objects are only loaded if within the bounds of the current viewable map

db/src/main/kotlin/edu/artic/db/daos/ArticMapAnnotationDao.kt
```diff
@@ -16,17 +17,24 @@ abstract class ArticMapAnnotationDao {
     @Query("select * from ArticMapAnnotation where annotationType = :type")
     abstract fun getAnnotationByType(type: String): Flowable<List<ArticMapAnnotation>>
 
-    @Query("select * from ArticMapAnnotation where annotationType = \"Text\" and textType = :type")
+    @Query("select * from ArticMapAnnotation where annotationType = :type and floor = :floor")
+    abstract fun getAnnotationByTypeForFloor(type: String, floor: String): Flowable<List<ArticMapAnnotation>>
+
+    @Query("select * from ArticMapAnnotation where annotationType = \"Text\" and textType = :type ")
```
agrosner: for `Text` can you reuse `annotationType` like so:
```
 @Query("select * from ArticMapAnnotation where annotationType = :annotationType and textType = :type ")
   abstract fun getTextAnnotationByType(type: String, annotationType: String = ArticMapAnnotationType.TEXT): Flowable<List<ArticMapAnnotation>>
```

db/src/main/kotlin/edu/artic/db/daos/ArticMapAnnotationDao.kt
```diff
@@ -16,17 +17,24 @@ abstract class ArticMapAnnotationDao {
...
     abstract fun getTextAnnotationByType(type: String): Flowable<List<ArticMapAnnotation>>
 
+    @Query("select * from ArticMapAnnotation where annotationType = \"Text\" and textType = :type and floor = :floor")
```
agrosner: same here as above

db/src/main/kotlin/edu/artic/db/daos/ArticMapAnnotationDao.kt
```diff
@@ -16,17 +17,24 @@ abstract class ArticMapAnnotationDao {
...
+    abstract fun getTextAnnotationByTypeAndFloor(type: String, floor: String): Flowable<List<ArticMapAnnotation>>
+
     fun getBuildingNamesOnMap(): Flowable<List<ArticMapAnnotation>> {
         return getAnnotationByType("Text")
     }
-    fun getAmenitiesOnMap() : Flowable<List<ArticMapAnnotation>> {
-        return getAnnotationByType("Amenity")
+
+    fun getAmenitiesOnMapForFloor(floor: String): Flowable<List<ArticMapAnnotation>> {
+        return getAnnotationByTypeForFloor("Amenity", floor)
```
agrosner: for single-line methods I prefer expression bodies:
```
fun getAmenitiesOnMapForFloor(floor: String): Flowable<List<ArticMapAnnotation>> = getAnnotationByTypeForFloor("Amenity", floor)
```

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -250,9 +265,40 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
         mapView.onLowMemory()
     }
 
+    private fun loadAmenities(amenitiesList: List<MapItem.Annotation>) {
+        amenitiesMarkerList
+                .forEach { marker -> marker.remove() }
+
+        amenitiesList.forEach { mapItem ->
```
agrosner: looks like you're re-adding the markers but not removing them physically from the list.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -250,9 +265,40 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
...
+                            .position(mapItem.item.toLatLng())
+                            .zIndex(0f)
+                    )
+
+            )
+        }
+    }
+
+    private fun loadSpaceOrLandmarks(spaceOrLandmarkList: List<MapItem.Annotation>) {
+        spaceOrLandmarkMarkerList
```
agrosner: same here.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -250,9 +265,40 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
...
+                .forEach { marker -> marker.remove() }
+
+        spaceOrLandmarkList.forEach { mapItem ->
```
agrosner: Ive seen this code repeated in multiple places doing the exact same thing save a small change in the `BitmapDescriptorFactory`. please consider creating an intermediate `inline` method that does all the logic for galleryMarkers, spaceOrLandmarkerList, etc. the same way:
```
private inline fun loadMarkers(list: List<MapItem.Annotation>, markerBuilder: (MapItem.Annotation) -> Marker) {
  list.forEach {  it.remove () }
  list.clear()
  list.add(map.addMarker(markerBuilder(mapItem)))
}
```

map/src/main/kotlin/edu/artic/map/MapViewModel.kt
```diff
@@ -56,22 +57,35 @@ class MapViewModel @Inject constructor(
...
+                floor.distinctUntilChanged()
+        ) { _, floor ->
+            mapAnnotationDao.getTextAnnotationByTypeAndFloor(
+                    ArticMapTextType.SPACE,
+                    floor.toString())
+                    .asObservable()
+                    .map { it.mapToMapItem() }
+        }
+                .flatMap {
+                    it
```
agrosner: don't think this is needed now?

raigex: yea it is. for some reason otherwise it returns an observable<...> this is the only way I can think of to unwrap it @agrosner

map/src/main/kotlin/edu/artic/map/MapFragment.kt
raigex: I guess there can be an inline method for this...kinda

map/src/main/kotlin/edu/artic/map/MapFragment.kt
raigex: fixed.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
raigex: fixed

map/src/main/kotlin/edu/artic/map/MapFragment.kt
raigex: Created it and used where it makes sense (basically amentities and landmarkAndSpace

db/src/main/kotlin/edu/artic/db/daos/ArticMapAnnotationDao.kt
raigex: So use this if we want default?

PR #36, opened 2018-08-09
[Documentation] - add an explicit AGPL v3 license

Cliabhach:
This file matches that used by the ios codebase.


Cliabhach:
@raigex Would you be willing to take a look at this?

PR #41, opened 2018-08-09
[CI] - ensure SDK 27 is installed for CircleCI

Cliabhach:
The existing config file erroneously requests an image with SDK 25 instead. Perhaps this will fix the failures we've been seeing?


Cliabhach:
Well, it _looks_ good, I guess. Will need to fix that `store_artifacts` to be a bit smarter (i.e. detect non-app reports) in future.

PR #42, opened 2018-08-10
AIC-51 Bug Fixes

raigex:


map/src/main/kotlin/edu/artic/map/MapViewModel.kt
```diff
@@ -157,11 +158,23 @@ class MapViewModel @Inject constructor(
                 galleries,
                 objects
         ) { floor, galleryList, objectList ->
-            val list = mutableListOf<MapItem<*>>()
-            list.addAll(galleryList.map { gallery -> MapItem.Gallery(gallery, floor) })
-            list.addAll(objectList.map { articObject -> MapItem.Object(articObject, floor) })
-            return@combineLatest list
-        }
+            return@combineLatest if (currentZoomLevel == MapZoomLevel.Three && floor == currentFloor) {
```
agrosner: Why do this check then do `filter { it.isNotEmpty() }`. Unneccessary. Just do `filter { floor, _, _ -> (currentZoomLevel == MapZoomLevel.Three && floor == currentFloor) }` initially on the chain, then do your logic next.

agrosner: also shouldn't current zoom level be an `Observable`? weird to have it here as just a variable.

raigex: it's more so that we can get it properly set up in the fragment level without having to call has value and get value of floor observable when loading from glide

PR #45, opened 2018-08-14
Fix map crash on motorola devices (AIC-261)

Cliabhach:
Also includes some documentation updates.


Cliabhach:
@agrosner I'm just going to rebase on dev

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -247,6 +267,26 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
         }
     }
 
+    @UiThread
+    protected fun deriveBaseOverlayDescriptor(host: Activity): BitmapDescriptor {
```
agrosner: This should go in a helper method somewhere for the map image. maybe an `ImageExtensions` class?.
Also use lowest denominator, in this case swap `Activity` with `Context`.

base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
```diff
@@ -17,3 +19,57 @@ fun Activity.setWindowFlag(bits: Int, on: Boolean) {
...
+ * and the remaining indicators are not checked. They are as follows:
+ * 1. is this object [not allowed to access the filesystem][Context.isRestricted]?
+ * 1. are we running on a [low-RAM device][ActivityManager.isLowRamDevice]?
+ * 1. do we only have access to [one processor][Runtime.availableProcessors]?
+ * 1. is the maximum memory available to our process under [175 megabytes][MIN_MEMORY_FOR_HIGH_QUALITY]?
+ *
+ * If you're wondering about that last check, the amount of RAM we'd prefer to
+ * use is noticeably higher than AOSP's 'low RAM' threshold.
+ */
+fun Context.isResourceConstrained() : Boolean {
```
agrosner: nice! this will work well for now until we get tiling working. Even then there might be some optimizations we can do in other places.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
Cliabhach: (for anyone watching, @agrosner confirmed that offline)

Alright, the new commits have that change. Hopefully it's all good.

PR #47, opened 2018-08-14
Playback Analytics & Notification Icons

sam33rdhakal:
updates
* Adds audio completed, interrupted analytics events
* Adds notification icons


db/src/main/kotlin/edu/artic/db/models/ModelExtensions.kt
```diff
@@ -0,0 +1,13 @@
+package edu.artic.db.models
+
+/**
+ * @author Sameer Dhakal (Fuzz)
+ */
+
+fun ArticObject.getAudio(): ArticAudioFile? {
+    return if (this.audioCommentary.isNotEmpty()) {
+        this.audioCommentary.first().audioFile
```
agrosner: we could cleanup this method to:
```
fun ArticObject.getAudio(): ArticAudioFile? = this.audioCommentary.firstOrNull()?.audioFile
```

agrosner: also renaming it to an Extension property might be better:
```
val ArticObject.audioFile: ArticAudioFile? 
  get() = this.audioCommentary.firstOrNull()?.audioFile
```
And more semantic to Kotlin, not Java "getters".

db/src/main/kotlin/edu/artic/db/models/ModelExtensions.kt
```diff
@@ -0,0 +1,13 @@
+package edu.artic.db.models
```
agrosner: Rename file (too generic) or put it in the same file as `ArticObject` as an extension method residing near the class.

media/src/main/java/edu/artic/media/audio/AudioPlayerService.kt
```diff
@@ -144,16 +150,24 @@ class AudioPlayerService : Service() {
 
     val disposeBag = DisposeBag()
 
+    @Inject
+    lateinit var analyticsTracker: AnalyticsTracker
+
     override fun onCreate() {
+        AndroidInjection.inject(this)
```
agrosner: There is a `DaggerService` that does this for you, if you prefer.

sam33rdhakal: Updated.

db/src/main/kotlin/edu/artic/db/models/ModelExtensions.kt
sam33rdhakal: Removed this file and moved the extension method to ArticObject as the extension property.

db/src/main/kotlin/edu/artic/db/models/ModelExtensions.kt
sam33rdhakal: I implemented extension property.

PR #48, opened 2018-08-14
Events Date issue

sam33rdhakal:
Fixes date time issue.


base/src/main/java/edu/artic/base/utils/DateTimeHelper.kt
```diff
@@ -27,4 +29,8 @@ class DateTimeHelper {
                 .appendPattern("MMMM d   h:mm a")
                 .toFormatter()
     }
-}
\ No newline at end of file
+}
+
+fun ZonedDateTime.getLocalDateTime(): ZonedDateTime {
```
agrosner: This is not returning a `LocalDateTime` which means its confusing. You mean `convertToLocalTime`?

db/src/main/kotlin/edu/artic/db/AppConverters.kt
```diff
@@ -149,12 +147,12 @@ class AppConverters {
...
     @TypeConverter
-    fun fromDateTime(localDateTime: LocalDateTime?): Long? = localDateTime?.let {
-        it.toInstant(ZoneId.systemDefault().rules.getOffset(it)).toEpochMilli()
+    fun fromDateTime(input: ZonedDateTime?): String? = input?.let {
+        it.toString()
     }
 
     @TypeConverter
-    fun toDateTime(time: Long?): LocalDateTime? = time?.let { LocalDateTime.ofInstant(Instant.ofEpochMilli(it), ZoneId.systemDefault()) }
+    fun toDateTime(time: String?): ZonedDateTime? = time?.let { DateTimeFormatter.ISO_DATE_TIME.parse(it, ZonedDateTime.FROM) }
```
agrosner: might be worth doing a `try..catch datetimeparseexception` I think it throws? but we should also just log it if it does, but to be safe we dont crash here.

db/src/main/kotlin/edu/artic/db/models/ArticEvent.kt
```diff
@@ -4,21 +4,31 @@ import android.arch.persistence.room.Entity
...
         @Json(name = "location") val location: String?,
         @Json(name = "id") @PrimaryKey val id: Int,
         @Json(name = "button_text") val button_text: String?,
         @Json(name = "title") val title: String,
-        @Json(name = "start_at") val start_at: LocalDateTime
+        @Json(name = "start_at") val start_at: ZonedDateTime
 
-) : Parcelable
+) : Parcelable {
+    fun getStartTime(): ZonedDateTime {
```
agrosner: property syntax :) please lets move away from getter/setter thinking and more property thinking.

db/src/main/kotlin/edu/artic/db/models/ArticEvent.kt
```diff
@@ -4,21 +4,31 @@ import android.arch.persistence.room.Entity
...
+        return start_at.getLocalDateTime()
+    }
+
+    fun getEndTime(): ZonedDateTime {
```
agrosner: same here as above.

db/src/main/kotlin/edu/artic/db/models/ArticExhibition.kt
```diff
@@ -4,20 +4,30 @@ import android.arch.persistence.room.Entity
...
         @Json(name = "gallery_id") val gallery_id: String?,
         @Json(name = "id") @PrimaryKey val id: Int,
-        @Json(name = "aic_end_at") val aic_end_at: LocalDateTime,
+        @Json(name = "aic_end_at") val aic_end_at: ZonedDateTime,
         @Json(name = "title") val title: String,
         var order: Int = -1
-) : Parcelable
\ No newline at end of file
+) : Parcelable {
+    fun getStartTime(): ZonedDateTime {
```
agrosner: same as above

db/src/main/kotlin/edu/artic/db/models/ArticExhibition.kt
```diff
@@ -4,20 +4,30 @@ import android.arch.persistence.room.Entity
...
+        return aic_start_at.getLocalDateTime()
+    }
+
+    fun getEndTime(): ZonedDateTime {
```
agrosner: reminder same as above.

db/src/main/kotlin/edu/artic/db/AppConverters.kt
sam33rdhakal: I will check the date validity in Moshi adapter. If we catch the invalid dates when we are serializing we don't need it here. Right ?

base/src/main/java/edu/artic/base/utils/DateTimeHelper.kt
sam33rdhakal: We are storing datetime in `UTC format` but need to convert it to the user's timezone before we dispaly. This is handy extension function to add updated timezone info to ZonedDateTime. Since it was confusing, I renamed it to `toCurrentTimeZone`.

db/src/main/kotlin/edu/artic/db/models/ArticExhibition.kt
sam33rdhakal: Done.

db/src/main/kotlin/edu/artic/db/models/ArticExhibition.kt
sam33rdhakal: Done.

db/src/main/kotlin/edu/artic/db/models/ArticEvent.kt
sam33rdhakal: Done.

db/src/main/kotlin/edu/artic/db/models/ArticEvent.kt
sam33rdhakal: Updated. :)

PR #49, opened 2018-08-14
Map department pins are transient (AIC-251)

Cliabhach:
This includes some much-wanted documentation of the `MapViewModel` class, as well as a fix for the inconsistent loading issue. As a bonus, we can now use the usual channels to tell the fragment that our map needs to be cleared of all markers.

### DIFF
--- a/db/src/main/kotlin/edu/artic/db/models/ArticGallery.kt
+++ b/db/src/main/kotlin/edu/artic/db/models/ArticGallery.kt
@@ -16,6 +16,12 @@ data class ArticGallery(
         @Json(name = "latitude" ) val latitude: Double,
         @Json(name = "longitude" ) val longitude: Double,
         @Json(name = "floor" ) val floor: String?,
+        /**
+         * NB: as established by the iOS codebase, please use [title] instead of this field.
+         *
+         * In the general sense we expect the two fields to be equivalent and this one may
+         * be removed from the dta model in a future commit.
+         */
         @Json(name = "title_t" ) val titleT: String?,
         @Json(name = "gallery_id" ) val galleryId: String?,
         @Json(name = "is_boosted" ) val isBoosted: Boolean,
--- a/db/src/main/kotlin/edu/artic/db/models/ArticGallery.kt
+++ b/db/src/main/kotlin/edu/artic/db/models/ArticGallery.kt
@@ -23,4 +23,10 @@ data class ArticGallery(
         @Json(name = "closed" ) val closed: Boolean,
         @Json(name = "number" ) val number: String?,
         @Json(name = "category_titles" ) val categoryTitles: List<String>
-)
+) {
+    /**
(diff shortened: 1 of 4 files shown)


db/src/main/kotlin/edu/artic/db/models/ArticGallery.kt
```diff
@@ -23,4 +23,10 @@ data class ArticGallery(
         @Json(name = "closed" ) val closed: Boolean,
         @Json(name = "number" ) val number: String?,
         @Json(name = "category_titles" ) val categoryTitles: List<String>
-)
\ No newline at end of file
+) {
+    /**
+     * Returns [floor], parsed to an integer. We default to [Int.MIN_VALUE] as 0 is a valid floor.
+     */
+    val floorAsInt: Int
```
agrosner: perfect

map/src/main/kotlin/edu/artic/map/MapViewModel.kt
```diff
@@ -116,72 +149,86 @@ class MapViewModel @Inject constructor(
...
+     *
+     * Each of the observed [MapItem.Object]s include info about the floor
+     * of the gallery where it is found.
+     */
+    fun observeObjectsWithin(observed: Observable<List<ArticGallery>>): Observable<List<MapItem.Object>> {
+        return observed.map { galleries ->
+            galleries.filter { gallery ->
+                gallery.titleT != null
+            }.map { gallery ->
+                val title = gallery.titleT.orEmpty()
```
agrosner: switch to `title` not `titleT` for its title

map/src/main/kotlin/edu/artic/map/MapViewModel.kt
```diff
@@ -116,72 +149,86 @@ class MapViewModel @Inject constructor(
...
+     * we use to retrieve these objects precludes the possibility of
+     * duplicates.
+     *
+     * Each of the observed [MapItem.Object]s include info about the floor
+     * of the gallery where it is found.
+     */
+    fun observeObjectsWithin(observed: Observable<List<ArticGallery>>): Observable<List<MapItem.Object>> {
+        return observed.map { galleries ->
+            galleries.filter { gallery ->
+                gallery.titleT != null
```
agrosner: here too.

map/src/main/kotlin/edu/artic/map/MapViewModel.kt
Cliabhach: Hmm. I _was_ just preserving the previous behavior, but it doesn't look like we use `titleT` anywhere but the two places you flagged up. Let me confirm that there wouldn't be any adverse effects.

Cliabhach: Right, I've made the change. The new commit (174dcfca339c741c10) should resolve that.

PR #54, opened 2018-08-19
Welcome Fragment / Details Headers Enhancements

agrosner:
Originally this was to use `MotionLayout` as replacement for `CollapsingToolbarLayout`, but this MR instead will focus on flattening the view hierarchy.
1. Flatten nested `LinearLayout` inside `ConstraintLayout` with direct constraints.
2. Add press states to "See All" buttons
3. Consolidate `AppBarLayout` for both `Welcome` and `Information` screens into same layout.
4. Better previewing of the `WelcomeFragment`.
5. Consolidate headers of OnView, Events, and Tour Details into the same, reusable component. Hopefully we can consolidate even more into the same layout.


Cliabhach:
Well, this is a pleasant surprise. Two things come to mind right away:

Commit bd85943 removes a file from .idea and removes a line from `audio/build.gradle`. Both of these are outside the scope of the declared merge request and should be filed separately. Failing that, they should at least be placed in a separate commit.

I'm not a big fan of your commit messages for cee51f35 and 2796e5f. We generally use `[Refactor]`, not `[gardening]`, to refer to changes that do not materially affect behavior. From viewing the code I'd say these

1. improve indentation in WelcomeFragment.kt
and
2. remove an unnecessary style

In both cases, "code cleanup from some files" is excessively vague.

agrosner:
@Cliabhach 👍 thanks for your input
1. The removing line from `audio/build.gradle` yes it is out of the scope, however it consolidates the `ConstraintLayout` version that I set in the main project. It is extra line that should be removed. Removing .idea files, is there a reason why we've checked them into the repo in the first place? Never sought a real neeed...
2. yes I guess it is vague. I will make better commit messages in the future. Gardening is what iOS uses to mark commits that are about cleaning up code or updating dependencies. I could use refactor, but I think it is something that should be unified.

agrosner:
@sam33rdhakal please review

PR #59, opened 2018-08-22
[Glide] reduce error verbosity

Cliabhach:
In release builds, glide will no longer print lots of stacktraces to logcat. Also:

* reduced memory usage when rendering a non-square image
* progressive loading of thumbnail then full image


Cliabhach:
Some of the image urls for the thumbnails return an error even though the full url is perfectly functional.

PR #65, opened 2018-08-24
Crash Fix

sam33rdhakal:
Fixes the application crash when user taps start tour button. Basically, the issue was floor was emitting the garbage value.

### DIFF
--- a/map/src/main/kotlin/edu/artic/map/MapFragment.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapFragment.kt
@@ -348,7 +348,7 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
                 groundOverlayGenerated.filter { it }
         ) { floor, _ ->
             floor
-        }.filter { floor -> floor > -1 }
+        }.filter { floor -> floor > Int.MIN_VALUE }
                 .subscribe {
                     buildingGroundOverlay.setImage(BitmapDescriptorFactory.fromAsset("AIC_Floor$it.png"))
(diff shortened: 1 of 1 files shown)


map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -352,9 +348,10 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
                 groundOverlayGenerated.filter { it }
         ) { floor, _ ->
             floor
-        }.subscribe {
-            buildingGroundOverlay.setImage(BitmapDescriptorFactory.fromAsset("AIC_Floor$it.png"))
-        }.disposedBy(disposeBag)
+        }.filter { floor -> floor > -1 }
```
Cliabhach: Please compare with Int.MIN_VALUE (the default used in ArticTour,etc.) instead.

sam33rdhakal: Updated.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -348,7 +348,7 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
                 groundOverlayGenerated.filter { it }
         ) { floor, _ ->
             floor
-        }.filter { floor -> floor > Int.MIN_VALUE }
+        }.filter { floor -> floor in 0..3 }
```
Cliabhach: Excellent, love that solution.

PR #67, opened 2018-08-24
Map: restrict which objects load

Cliabhach:
This greatly reduces the number of duplicate requests.

### DIFF
--- a/map/src/main/kotlin/edu/artic/map/MapActivity.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapActivity.kt
@@ -32,8 +32,6 @@ class MapActivity : BaseActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
-        // The map is resource-intensive. Be kind to it.
-        Glide.get(this).setMemoryCategory(MemoryCategory.LOW)
 
         bottomNavigation.apply {
             disableShiftMode(R.color.map_menu_color_list)
@@ -42,9 +40,16 @@ class MapActivity : BaseActivity() {
         }
     }
 
-    override fun onDestroy() {
-        super.onDestroy()
+    override fun onStart() {
+        super.onStart()
+        // The map is resource-intensive. Be kind to it.
+        Glide.get(this).setMemoryCategory(MemoryCategory.LOW)
+    }
 
+    override fun onStop() {
+        super.onStop()
+        // The GoogleMap itself will lower its usage at this point, so it is
+        // safe to increase the memory category too.
         Glide.get(this).setMemoryCategory(MemoryCategory.NORMAL)
     }
 
--- a/map/src/main/kotlin/edu/artic/map/MapActivity.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapActivity.kt
@@ -1,6 +1,8 @@
 package edu.artic.map
 
 import android.os.Bundle
+import com.bumptech.glide.Glide
+import com.bumptech.glide.MemoryCategory
 import edu.artic.base.utils.disableShiftMode
 import edu.artic.db.models.ArticTour
 import edu.artic.navigation.NavigationSelectListener
@@ -29,10 +31,21 @@ class MapActivity : BaseActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        // The map is resource-intensive. Be kind to it.
+        Glide.get(this).setMemoryCategory(MemoryCategory.LOW)
+
         bottomNavigation.apply {
             disableShiftMode(R.color.map_menu_color_list)
             selectedItemId = R.id.action_map
             setOnNavigationItemSelectedListener(NavigationSelectListener(this.context))
         }
     }
+
+    override fun onDestroy() {
+        super.onDestroy()
+
+        Glide.get(this).setMemoryCategory(MemoryCategory.NORMAL)
+    }
+
 }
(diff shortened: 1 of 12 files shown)


map/src/main/kotlin/edu/artic/map/MapActivity.kt
```diff
@@ -29,10 +31,21 @@ class MapActivity : BaseActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+
+        // The map is resource-intensive. Be kind to it.
+        Glide.get(this).setMemoryCategory(MemoryCategory.LOW)
```
agrosner: Move this into `onStart/onStop` since map might be around but on the backstack.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -515,17 +554,95 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
...
+            itemList.chunked(2 * coreCount)
+        } else {
+            // Work with 3 x 'number of processor cores' items at a time.
+            itemList.chunked(3 * coreCount)
+        }
+
+                Observables.zip(
+                        // There's an Observable operator 'window', but that's (severely) restricted to a single subscriber
+                        batchedItems.toObservable(),
+                        // The 'interval' here will make sure only one event is emitted every 100 ms
```
agrosner: just use `throttleLast()` with 100 MS. does it more cleanly.

agrosner: throttle last will also emit if theres more events within the 100 ms timeframe. this solution really just delays it by 100ms.

agrosner: http://reactivex.io/documentation/operators/sample.html

agrosner:
```
 Returns an Observable that emits only the last item emitted by the source Observable during sequential
     * time windows of a specified duration.
     * <p>
     * This differs from {@link #throttleFirst} in that this ticks along at a scheduled interval whereas
     * {@link #throttleFirst} does not tick, it just tracks passage of time.
```
This might work so it continues

agrosner: and increasing the delay to like to 500 ms might help. it will be slightly delayed but will pick up anything that happens within the 500ms and use the last even before emitting

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -515,17 +554,95 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
...
+                )
+                .map {
+                    val mapItems = it.first
+
+                    return@map mapItems.filter(this@MapFragment::alreadyHasMarker)
+                }.filter {
+                    it.isNotEmpty()
+                }.subscribeBy {
+                    it.toObservable()
+                            .observeOn(AndroidSchedulers.mainThread())
```
agrosner: Just call `observeOn` in the chain before `subscribeBy` then you can just call `bindtoMap` directly.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -738,4 +926,34 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
...
+private fun LatLng.isCloseEnoughToCenter(bounds: LatLngBounds): Boolean {
+    return bounds.contains(this) && bounds.center.distanceTo(this) < 15
+}
+
+/**
+ * Alias to [Location.distanceBetween], where 'this' is the first param and 'other' is the second.
+ *
+ * @return a distance, in meters
+ */
+private fun LatLng.distanceTo(other: LatLng): Float {
```
agrosner: move this to `MapUtils` method. no need to pollute the `MapFragment` class with extra logic especially if we end up needing it somewhere else. :)

map/src/main/kotlin/edu/artic/map/helpers/CollectionExtensions.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Simple inversion of [Iterable.contains].
+ */
+fun <T> Iterable<T>.doesNotContain(element: T): Boolean {
+    return !this.contains(element)
+}
+
+/**
+ * Copy of [java.util.Collection.removeIf]. Functionality is identical.
+ */
+fun <E> MutableCollection<E>.modifyThenRemoveIf(filter: (E) -> Boolean) : Boolean {
```
agrosner: theres no kotlin method to do this? theres a `filter` method which returns a list of filtered items matching the predicate.

agrosner: it returns a new list, but its better to deal with immutable lists in my opinion

map/src/main/kotlin/edu/artic/map/MapFragment.kt
Cliabhach: (we talked a little offline about the intent of this code)

The main goal is not so much to prevent too many events from being handled, but to slow down the rate at which we handle them. So `window` was very nearly what I needed. @agrosner communicated that he is working on a great improvement to the way `itemList` is passed in, so with an eye towards replacing it soon the code may stand.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
Cliabhach: Yeah, good call.

Cliabhach:
I'm dropping the wip and going to focus on fixing the issues @agrosner raised.

agrosner:
@Cliabhach take a look at #35 . It changes things pretty significantly. Its not worth further efforts on this as it will overlap. I will add in the `VisibleRegion` functionality on my branch too to render the marker "dots" when not in focus. But I guess a question of how necessary that specific functionality is. As it stands it seems to work well.

Cliabhach:
All right, we'll leave it as is. I'll rebase on dev and you can merge in.

agrosner:
sounds good.

Cliabhach:
Rebase complete, now testing before I push up....

Cliabhach:
It's pushed. @agrosner , it's all you.

PR #68, opened 2018-08-24
Do not use full-size images on map

Cliabhach:


Cliabhach:
Glad to see you still like it. I'll merge in shortly, then.

PR #69, opened 2018-08-24
Small Map cleanup

agrosner:
1. Clean up some of the verbose Kotlin code.
2. Extract conversion logic for `ArticMapAmenityType` into a package local helper method, rather than placing that in the Fragment.
3. Consolidate the different small generator classes into a single multi-class file, aiding in organization and removing need for inner package.


map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -9,22 +9,43 @@ import com.bumptech.glide.Glide
...
+import com.google.android.gms.maps.model.BitmapDescriptor
+import com.google.android.gms.maps.model.BitmapDescriptorFactory
+import com.google.android.gms.maps.model.CameraPosition
+import com.google.android.gms.maps.model.GroundOverlay
+import com.google.android.gms.maps.model.GroundOverlayOptions
+import com.google.android.gms.maps.model.LatLng
+import com.google.android.gms.maps.model.LatLngBounds
+import com.google.android.gms.maps.model.MapStyleOptions
+import com.google.android.gms.maps.model.Marker
+import com.google.android.gms.maps.model.MarkerOptions
```
Cliabhach: Something seems odd about your import settings - is the IDE automatically replacing the `*` with explicit imports?

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -298,46 +319,27 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
...
-                                R.drawable.map_floor_background_default
-                    )
-                    floorThree.setBackgroundResource(
-                            if (it == 3)
-                                R.drawable.map_floor_background_selected
-                            else
-                                R.drawable.map_floor_background_default
-                    )
+                .subscribeBy { floor: Int ->
+                    fun backgroundForState(whichFloor: Int): Int {
```
Cliabhach: Nice, I like this.

map/src/main/kotlin/edu/artic/map/MarkerGenerators.kt
```diff
@@ -0,0 +1,139 @@
...
+import android.graphics.Canvas
+import android.graphics.Color
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ImageView
+import android.widget.TextView
+
+/**
+ * Description:
```
Cliabhach: Drop the first comment block, it's not doing anything much.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
agrosner: yeah. It turned off `*` imports. Not sure if thats an issue, something to discuss later. I think for now ill keep it as is and we can tweak that setting in the project so its the same for everyone.

map/src/main/kotlin/edu/artic/map/MarkerGenerators.kt
agrosner: oops great catch! ha.

PR #71, opened 2018-08-26
WIP: Map refactor

agrosner:
This is a preview of the changes that are coming to the map. 
1. Adds new `image` module to handle glide extensions and glide in general. Before this was in the `base` and `welcome` modules??
2. Manually populate `floor` field into the `ArticObject` upon database save so we can directly reference this. Just as iOS does.
3. Adds `RXGlideExtensions` which provides an `Observable` for requests, so we can let RX manage the lifecycle.
4. Adds The concept of the `MapItemRenderer`. It's a common abstraction between all of the map markers to enable them to load their content dynamically on the map while map change event occurs based on `MapFocus`, Floor, and `MapDisplayMode`. 
5. Marker asynchronous rendering happens efficiently. We queue up all mapitem modifications on the `TrampolineScheduler` to execute these in order, without overlap. We provide a `Subject` which buffers new `MarkerHolder` objects based on 20 at time, or for once every second of sampling, whichever happens first. Then each batch of items are executed in order, 100ms apart in order to not overload/block the UI while we add them to the map. Once these all complete in order we remove any existing markers that are the same from the map (to allow us to essentially replace markers as this happens). If this is cut short, we utilize a temp marker list which gets flushed on next map event (such as focus, floor, or display mode changes) so the markers do not stick around on the map.
6. We have the ability to display a "progress" marker on the map before the asynchronous method is called, this is not implemented yet as we need an asset to display this state.
7. The extensible `MapItemRenderer` means we can reuse it across all implementations and add new ones if needed.


Cliabhach:
As a rule of thumb, if a commit message can only be expressed with multiple periods or `and`s it should be broken up. Pull #72 is now ready.

db/src/main/kotlin/edu/artic/db/models/ArticTour.kt
```diff
@@ -93,10 +93,10 @@ data class ArticTour(
     }
 
     /**
-     * Returns [floor], parsed to an integer. We default to [Int.MIN_VALUE] as 0 is a valid floor.
+     * Returns [floor], parsed to an integer. We default to 0 since some tours can be on multiple floors.
      */
     val floorAsInt: Int
-        get() = floor?.toIntOrNull() ?: Int.MIN_VALUE
+        get() = floor?.toIntOrNull() ?: 0
```
sam33rdhakal: Here we can't use 0. If the floor is empty iOS app assigns first stop's floor to the tour.

agrosner: noted. will change

db/src/main/kotlin/edu/artic/db/models/ArticObject.kt
```diff
@@ -13,49 +14,52 @@ import kotlinx.android.parcel.Parcelize
...
 //        @Json(name = "audio") val audio: List<String?>, Removed for now until json gets fixed as sometimes returns string othertimes object
-        @Json(name = "audio_transcript") val audioTranscript: String?,
-        @Json(name = "object_selector_number") val objectSelectorNumber: String?,
-        @Json(name = "object_selector_numbers") val objectSelectorNumbers: List<String?>
+        @Json(name = "audio_transcript") val audioTranscript: String? = null,
+        @Json(name = "object_selector_number") val objectSelectorNumber: String? = null,
+        @Json(name = "object_selector_numbers") val objectSelectorNumbers: List<String?> = listOf(),
+
+        // manually populated via DB save.
+        var floor: Int = Int.MIN_VALUE
```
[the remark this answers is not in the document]
agrosner: lets make this a separate PR. I am filing an open enhancement ticket now.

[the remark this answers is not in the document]
agrosner: #74

db/src/main/kotlin/edu/artic/db/models/ArticTour.kt
agrosner: removed

PR #73, opened 2018-08-29
Correct an image loading oversight in AudioPlayerService

Cliabhach:
This is a follow-up to PR #68.

After carefully reviewing the code in preparation for the audio player work, I realized that I did not address the use of large (*_not_* full-size, so this isn't a critical situation) images in the audio player notification. With this merge, the urls will be redirected properly towards our IIIF CDN.

I also took the opportunity to update some documentation around the meaning of `Playable`.

### DIFF
--- a/audio/src/main/java/edu/artic/audio/AudioDetailsFragment.kt
+++ b/audio/src/main/java/edu/artic/audio/AudioDetailsFragment.kt
@@ -26,7 +26,6 @@ import edu.artic.db.models.AudioFileModel
 import edu.artic.image.listenerAnimateSharedTransaction
 import edu.artic.media.audio.AudioPlayerService
 import edu.artic.media.refreshPlayBackState
-import edu.artic.ui.util.isNotEmpty
 import edu.artic.viewmodel.BaseViewModelFragment
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.rxkotlin.subscribeBy
--- a/audio/src/main/java/edu/artic/audio/AudioDetailsFragment.kt
+++ b/audio/src/main/java/edu/artic/audio/AudioDetailsFragment.kt
@@ -26,6 +26,7 @@ import edu.artic.db.models.AudioFileModel
 import edu.artic.image.listenerAnimateSharedTransaction
 import edu.artic.media.audio.AudioPlayerService
 import edu.artic.media.refreshPlayBackState
+import edu.artic.ui.util.isNotEmpty
 import edu.artic.viewmodel.BaseViewModelFragment
 import io.reactivex.android.schedulers.AndroidSchedulers
 import io.reactivex.rxkotlin.subscribeBy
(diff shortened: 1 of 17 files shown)


Cliabhach:
Hey, no complaints from CodeClimate! I like that.

db/src/main/kotlin/edu/artic/ui/util/UrlExtension.kt
```diff
@@ -16,4 +18,31 @@ fun String.asCDNUri() : Uri {
...
         else -> parsed
     }
+}
+
+/**
+ * Returns this if it is not null, or else [Uri.EMPTY] if it _is_ `null`.
+ *
+ * This method is inspired by [String.orEmpty].
+ */
+fun Uri?.orEmpty(): Uri {
```
agrosner: no null check here on it. It will always return `Uri.EMPTY`

db/src/main/kotlin/edu/artic/ui/util/UrlExtension.kt
```diff
@@ -16,4 +18,31 @@ fun String.asCDNUri() : Uri {
...
+fun Uri.isEmpty(): Boolean {
+    return this == Uri.EMPTY
+}
+
+/**
+ * Checks whether this is `!=` to [Uri.EMPTY].
+ *
+ * This method is inspired by [String.isNotEmpty].
+ */
+fun Uri.isNotEmpty(): Boolean {
```
agrosner: why not use expression body?
```
fun Uri.isNotEmpty(): Boolean = this != Uri.EMPTY
```

db/src/main/kotlin/edu/artic/db/Playable.kt
```diff
@@ -1,9 +1,35 @@
 package edu.artic.db
 
+import android.net.Uri
```
agrosner: using an android `Uri` in a viewmodel means we might have trouble testing it without robolectric tests.

Cliabhach: We benefit from strong typing in Uri that a `String` typealias couldn't offer. I'll change it to `okhttp3.HttpUrl`, as proposed offline.

audio/src/main/java/edu/artic/audio/AudioDetailsFragment.kt
```diff
@@ -26,6 +26,7 @@ import edu.artic.db.models.AudioFileModel
 import edu.artic.image.listenerAnimateSharedTransaction
 import edu.artic.media.audio.AudioPlayerService
 import edu.artic.media.refreshPlayBackState
+import edu.artic.ui.util.isNotEmpty
```
Cliabhach: Because the observable streams in `AudioDetailsFragment` do not specify types, the existing `String.isNotEmpty` calls are converted directly to `Uri.isNotEmpty`. This is thus the only line change needed to make it work.

db/src/main/kotlin/edu/artic/db/Playable.kt
Cliabhach: No, wait, we can just switch back to Strings. Should've thought of that earlier....

Cliabhach:
Incidentally, @nakkamarra was of great help in tracking down some of the inconsistencies. I would not have found our omission in the `welcome` module otherwise.

PR #77, opened 2018-08-30
Search base

sam33rdhakal:
Adds search module with basic search activity and default search suggestions fragment.


Cliabhach:
So, no real content, nothing but the basic structure, correct?

settings.gradle
```diff
@@ -1,4 +1,4 @@
-include ':app', ':image',
+include ':app', ':image', ':search',
```
Cliabhach: We should probably move these onto separate lines....not that there's much call to edit the file in the first place.

search/src/main/res/layout/fragment_default_search_suggestions.xml
```diff
@@ -0,0 +1,14 @@
...
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context=".DefaultSearchSuggestionsFragment">
+
+
+    <TextView
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:text="Default search suggestions" />
```
Cliabhach: This'll go into a R.string resource later, correct?

search/build.gradle
```diff
@@ -0,0 +1,21 @@
+apply plugin: 'kotlin-android-extensions'
+apply plugin: 'kotlin-kapt'
+
+dependencies {
+
+    implementation 'com.android.support.constraint:constraint-layout:1.1.2'
```
agrosner: use common dep in module pleasee, same as support lib. `libs.constraint_layout`

search/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+    implementation 'com.android.support:support-v4:27.1.1'
```
Cliabhach: This should use the same `libs` constant defined in the root directory's `build.gradle` file.

search/build.gradle
Cliabhach: Use `libs.constraint_layout` instead, please.

search/src/main/res/layout/fragment_default_search_suggestions.xml
sam33rdhakal: I am solving this on different pull req. This was used as the dummy element.

search/build.gradle
sam33rdhakal: Now uses the `constraint_layout` defines in the root `build.gradle`.

search/build.gradle
sam33rdhakal: Done.

settings.gradle
sam33rdhakal: Fixed it. All the modules are in separate lines now.

search/build.gradle
sam33rdhakal: Updated.

search/src/main/res/layout/fragment_default_search_suggestions.xml
Cliabhach: Cool, that's fine. I'll resolve this discussion.

PR #78, opened 2018-08-31
Statusbar color issue fix 

sam33rdhakal:
Updates
* Instead of getting the specific color we pull the color by attribute defined in active theme so that it is dynamic and applies correct color.
* Marks the fragments `AloofFragment` if they are not suppose to modify parent activity's theme

### DIFF
--- a/base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
+++ b/base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
@@ -90,12 +90,13 @@ val Activity.statusBarHeight: Int
 /**
  * Takes the array of color attributes ids and return array of colors.
  */
-fun Context.getThemeColors(colors: IntArray): Array<ColorStateList?> {
-    val found: Array<ColorStateList?>
+fun Context.getThemeColors(colors: IntArray): Array<ColorStateList> {
+    val found: Array<ColorStateList>
     var typedArray: TypedArray? = null
     try {
         typedArray = this.obtainStyledAttributes(colors)
         found = (0 until colors.size)
+                .mapNotNull { it }
                 .map { typedArray.getColorStateList(it) }
                 .toTypedArray()
     } finally {
--- a/base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
+++ b/base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
@@ -96,7 +96,7 @@ fun Context.getThemeColors(colors: IntArray): Array<ColorStateList?> {
     try {
         typedArray = this.obtainStyledAttributes(colors)
         found = arrayOfNulls(colors.size)
-        for (i in 0..colors.size) {
+        for (i in 0 until colors.size) {
             found[i] = typedArray.getColorStateList(i)
         }
     } finally {
--- a/base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
+++ b/base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
@@ -3,6 +3,8 @@ package edu.artic.base.utils
 import android.app.Activity
 import android.app.ActivityManager
 import android.content.Context
+import android.content.res.ColorStateList
+import android.content.res.TypedArray
 
 /**
  * @author Sameer Dhakal (Fuzz)
@@ -84,3 +86,21 @@ val Activity.statusBarHeight: Int
         }
         return statusBarHeight
     }
+
+/**
+ * Takes the array of color attributes ids and return array of colors.
+ */
+fun Context.getThemeColors(colors: IntArray): Array<ColorStateList?> {
+    val found: Array<ColorStateList?>
+    var typedArray: TypedArray? = null
+    try {
+        typedArray = this.obtainStyledAttributes(colors)
+        found = arrayOfNulls(colors.size)
+        for (i in 0..colors.size) {
+            found[i] = typedArray.getColorStateList(i)
+        }
+    } finally {
+        typedArray?.recycle()
+    }
+    return found
+}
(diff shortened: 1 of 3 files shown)


base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
```diff
@@ -83,4 +85,22 @@ val Activity.statusBarHeight: Int
...
+/**
+ * Takes the array of color attributes ids and return array of colors.
+ */
+fun Context.getThemeColors(colors: IntArray): Array<ColorStateList?> {
+    val found: Array<ColorStateList?>
+    var typedArray: TypedArray? = null
+    try {
+        typedArray = this.obtainStyledAttributes(colors)
+        found = arrayOfNulls(colors.size)
+        for (i in 0 until colors.size) {
```
agrosner: could you do `found = (0 until colors.size).map { typedArray.getColorStateList(it) }.toTypedArray()` ?
Also why is this necessary? Why not just set the status bar color to the kind we want?

ui/src/main/kotlin/edu/artic/ui/BaseFragment.kt
```diff
@@ -123,4 +129,9 @@ abstract class BaseFragment : Fragment() {
...
 
 
-}
\ No newline at end of file
+}
+
+/**
+ * Marker interface to mark those fragments which shouldn't modify the parent activity theme.
+ */
+interface AloofFragment
```
agrosner: Give it a better name. or better yet, just use an open `val` that they can override to control this behavior.

Cliabhach: Oh, good idea. `open val mayOverrideStatusBarColor` or something like that?

base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
Cliabhach: We don't know what color to use until the fragment is added.

ui/src/main/kotlin/edu/artic/ui/BaseFragment.kt
agrosner: yeah to follow convention in other parts of it

base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
sam33rdhakal: @agrosner Updated the code. @Cliabhach is right for the second point.

ui/src/main/kotlin/edu/artic/ui/BaseFragment.kt
sam33rdhakal: Replaced the marker interface by parent class property.

ui/src/main/kotlin/edu/artic/ui/BaseFragment.kt
```diff
@@ -93,11 +93,16 @@ abstract class BaseFragment : Fragment() {
...
-            requireActivity().setWindowFlag(WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS, false)
-            requireActivity().window?.statusBarColor = Color.TRANSPARENT
-        } else {
-            requireActivity().window?.statusBarColor = ContextCompat.getColor(requireContext(), R.color.colorPrimary)
+        if (this !is AloofFragment) {
+            if (hasTransparentStatusBar()) {
+                requireActivity().setWindowFlag(WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS, false)
+                requireActivity().window?.statusBarColor = Color.TRANSPARENT
+            } else {
+                val primaryDarkColor = intArrayOf(android.support.design.R.attr.colorPrimaryDark)
```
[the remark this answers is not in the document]
sam33rdhakal: Leaving the code to use colorPrimaryDark for now. We could refactor it later in different merge.

base/src/main/java/edu/artic/base/utils/ActivityExtensions.kt
```diff
@@ -83,4 +85,21 @@ val Activity.statusBarHeight: Int
...
+/**
+ * Takes the array of color attributes ids and return array of colors.
+ */
+fun Context.getThemeColors(colors: IntArray): Array<ColorStateList?> {
+    val found: Array<ColorStateList?>
+    var typedArray: TypedArray? = null
+    try {
+        typedArray = this.obtainStyledAttributes(colors)
+        found = (0 until colors.size)
+                .map { typedArray.getColorStateList(it) }
```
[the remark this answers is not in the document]
sam33rdhakal: Makes sense. Updated.

ui/src/main/kotlin/edu/artic/ui/BaseFragment.kt
```diff
@@ -93,11 +99,16 @@ abstract class BaseFragment : Fragment() {
...
-            requireActivity().window?.statusBarColor = Color.TRANSPARENT
-        } else {
-            requireActivity().window?.statusBarColor = ContextCompat.getColor(requireContext(), R.color.colorPrimary)
+        if (overrideStatusBarColor) {
+            if (hasTransparentStatusBar()) {
+                requireActivity().setWindowFlag(WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS, false)
+                requireActivity().window?.statusBarColor = Color.TRANSPARENT
+            } else {
+                val primaryDarkColor = intArrayOf(android.support.design.R.attr.colorPrimaryDark)
+                requireContext().getThemeColors(primaryDarkColor)[0]?.defaultColor?.let {
```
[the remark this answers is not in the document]
sam33rdhakal: Updated.

Cliabhach:
Ah, but first I must rebase....

PR #79, opened 2018-08-31
Floor serialization

### LINKED ISSUE #74: Add Custom Floor Adapter for Moshi Parsing.
We can parse the floor out of the gallery directly.
Objects have a gallery location which has floor in it.
Adding floor var with Int.MIN_VALUE might lead to confusion/bug if the floor is not assigned before it is accessed.

We could add the moshi adapter to convert floor string into the int in ArticGallery.
You could use it in ArticGallery.

```
/**
 * Parses floor as an integer. We default to [Int.MIN_VALUE] as 0 is a valid floor.
 */
class FloorAdapter {

    @ToJson
    fun toText(@Floor floor: Int): String = floor.toString()

    @FromJson
    @Floor
    fun fromText(text: String?): Int {
        return if (text != null) {
            try {
                return text.toInt()
            } catch (e: NumberFormatException) {
                if (text == "LL") {
                    0
                } else {
                    INVALID_FLOOR
                }
            }
        } else {
            INVALID_FLOOR
        }

    }
}

@Retention(AnnotationRetention.RUNTIME)
@JsonQualifier
annotation class Floor
then annotate the floor field in data class

        @Floor @Json(name = "audio_transcript")  val floor: Int

```

sam33rdhakal:
* resolves #74 
* converts string floor to int in gallery and tour

### DIFF
--- a/db/src/main/kotlin/edu/artic/db/Constants.kt
+++ b/db/src/main/kotlin/edu/artic/db/Constants.kt
@@ -1,7 +1,12 @@
 package edu.artic.db
 
 /**
+ * This file contains the constant values used in the models.
  * @author Sameer Dhakal (Fuzz)
  */
 
+/**
+ * [INVALID_FLOOR] represents the invalid floor number.
+ * Since negative numbers are valid number for underground floors we are using Int.MIN_VALUE.
+ */
 const val INVALID_FLOOR: Int = Int.MIN_VALUE
--- a/db/src/main/kotlin/edu/artic/db/Constants.kt
+++ b/db/src/main/kotlin/edu/artic/db/Constants.kt
@@ -1,0 +1,7 @@
+package edu.artic.db
+
+/**
+ * @author Sameer Dhakal (Fuzz)
+ */
+
+const val INVALID_FLOOR: Int = Int.MIN_VALUE
--- a/db/src/main/kotlin/edu/artic/db/MoshiAdapters.kt
+++ b/db/src/main/kotlin/edu/artic/db/MoshiAdapters.kt
@@ -2,6 +2,7 @@ package edu.artic.db
 
 import android.support.annotation.Nullable
 import com.squareup.moshi.FromJson
+import com.squareup.moshi.JsonQualifier
 import com.squareup.moshi.Moshi
 import com.squareup.moshi.ToJson
 import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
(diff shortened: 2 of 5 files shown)


db/src/main/kotlin/edu/artic/db/Constants.kt
```diff
@@ -0,0 +1,7 @@
+package edu.artic.db
+
+/**
+ * @author Sameer Dhakal (Fuzz)
+ */
+
+const val INVALID_FLOOR: Int = Int.MIN_VALUE
```
agrosner: oh wait please add doc here on what this is. `Constants.kt` is a little too vague.

db/src/main/kotlin/edu/artic/db/MoshiAdapters.kt
```diff
@@ -63,4 +65,36 @@ class ZonedDateTimeAdapter {
 
     @FromJson
     fun fromText(text: String): ZonedDateTime = ZonedDateTime.parse(text, formatter)
-}
\ No newline at end of file
+}
+
+/**
+ * Parses floor as an integer. We default to [Int.MIN_VALUE] as 0 is a valid floor.
```
Cliabhach: Update text since we now have an `INVALID_FLOOR` constant. Maybe just add a blank line, then `@see INVALID_FLOOR`.

db/src/main/kotlin/edu/artic/db/MoshiAdapters.kt
```diff
@@ -63,4 +65,36 @@ class ZonedDateTimeAdapter {
...
+    fun toText(@Floor floor: Int): String = floor.toString()
+
+    @FromJson
+    @Floor
+    fun fromText(text: String?): Int {
+        return if (text != null) {
+            try {
+                return text.toInt()
+            } catch (e: NumberFormatException) {
+                if (text == "LL") {
```
Cliabhach: Explicit comment above this line: `"LL" stands for 'Lower level'` or something like that.

db/src/main/kotlin/edu/artic/db/Constants.kt
sam33rdhakal: Added.

db/src/main/kotlin/edu/artic/db/MoshiAdapters.kt
sam33rdhakal: Updated.

db/src/main/kotlin/edu/artic/db/MoshiAdapters.kt
sam33rdhakal: Modified the comment.

PR #81, opened 2018-08-31
Update documentation around custom adapters

Cliabhach:
There were some omissions in the description of these classes. Note that `BaseRecyclerViewAdapter` was added during the 'Tours Summary' PR, #4 . We'll need to update the licenses on these files as well to reflect their provenance - that shall be a separate PR, some time next week.

### DIFF
--- a/adapter/src/main/java/edu/artic/adapter/BaseRecyclerViewAdapter.kt
+++ b/adapter/src/main/java/edu/artic/adapter/BaseRecyclerViewAdapter.kt
@@ -6,6 +6,7 @@ import android.arch.paging.AsyncPagedListDiffer
 import android.arch.paging.PagedList
 import android.content.Context
 import android.support.annotation.LayoutRes
+import android.support.annotation.UiThread
 import android.support.v7.recyclerview.extensions.AsyncDifferConfig
 import android.support.v7.util.DiffUtil
 import android.support.v7.util.ListUpdateCallback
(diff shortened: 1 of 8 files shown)


adapter/src/main/java/edu/artic/adapter/BaseRecyclerViewAdapter.kt
```diff
@@ -37,12 +39,11 @@ abstract class BaseRecyclerViewAdapter<TModel, VH : BaseViewHolder>(
     interface OnItemClickListener<in T> {
 
         /**
-         * The item clicked.
-
-         * @param position   The position in the itemslist that was clicked.
-         * *
+         * Callback for [on-click listeners][View.setOnClickListener]. Always invoked
+         * on main thread.
```
agrosner: "Always invoked on main thread."
Isn't this always a given?

adapter/src/main/java/edu/artic/adapter/BaseRecyclerViewAdapter.kt
```diff
@@ -325,25 +346,32 @@ abstract class BaseRecyclerViewAdapter<TModel, VH : BaseViewHolder>(
...
-
-     * @param position index into the item list of the item whose ViewHolder type is
-     * *                 desired
-     * *
-     * @return that type
+     * Please use [getLayoutResId] instead. If two [BaseViewHolder]s possess
+     * the same layout but require different binding logic, isolate that
+     * distinction to the 3-args `onBindViewHolder(VH, TModel?, Int)`.
      */
+    @Deprecated("Implementors should not need to override this method; please migrate to 'getLayoutResId'.", ReplaceWith("getLayoutResId(position)"))
```
agrosner: nice! yes. I think you used the smart kotlin replace annotation too!

adapter/src/main/java/edu/artic/adapter/BaseRecyclerViewAdapter.kt
Cliabhach: Honestly, I was trying to follow some of the framework docs - they'll include a small sentence like that on listener interfaces.

adapter/src/main/java/edu/artic/adapter/BaseRecyclerViewAdapter.kt
Cliabhach: Aw, thanks - I did use that

Cliabhach:
@agrosner you'll need to re-approve (just rebased, nothing else changed)

PR #82, opened 2018-09-03
Simplify access to AppDataObject

Cliabhach:
This is primarily a documentation update to the `:db` module.

I have made a field named `dataObject` in AppDataManager.kt, in the hopes that it will be easier for future developers to note this important field and make use of it directly. In the ideal case, this (or a similar change) could be somehow injected into `String.asCDNUri` - @agrosner, I defer to your experience on the matter. If you consider the field unnecessary I'll remove it from this PR and we will solve the image-url issue separately.


Cliabhach:
Ugh, the build failed with `Error: Unexpected failure during lint analysis of AudioDetailsFragment.kt (this is a bug in lint or one of the libraries it depends on)`, with another message (`java.io.IOException: Cannot allocate memory`) hinting at memory limitations. We'll try again later.

db/src/main/kotlin/edu/artic/db/RetrofitAppDataServiceProvider.kt
```diff
@@ -22,6 +29,11 @@ class RetrofitAppDataServiceProvider(
         const val EVENT_HEADER_ID = "events_download_header_id"
     }
 
+    /**
+     * This is used to access the correct 'events' and 'exhibitions' urls.
+     *
+     * Similar concept in play as with [AppDataManager.dataObject].
+     */
     val dataObject: Subject<ArticDataObject> = BehaviorSubject.create()
```
agrosner: This should be removed and implementation should be replaced with `AppDataManager` class. I do not think this belongs here anymore then or at least make this `private` so other classes can't see it.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -39,7 +50,49 @@ class AppDataManager @Inject constructor(
...
-     * @return an observable that will show how close to completion we are and a Done if fully complete
+     * In-memory cache of the [ArticDataObject] currently backing
+     * this app.
+     *
+     * Use it to access important network communication config like
+     * [ArticDataObject.imageServerUrl] and [ArticDataObject.dataApiUrl].
+     *
+     * Values are sourced from [the ArticAppData][ArticAppData.data]
+     * parsed in [getBlob]. Be advised that if no such value has been
+     * yet found in the time since first install, this will provide only
```
agrosner: no nulls allowed in RXJava, so this will _not_ emit.

agrosner: so either we make this `Optional<ArticDataObject>` or we're ok with registered listeners not emitting.

Cliabhach: Hmm. I think we'd be ok with it not emitting. What I would like you to advise on is if we could somehow inject this data into the sundry DAOs so that the object is available to the `asCDNUri` function. I don't see how this could fail to emit in such a case.

db/src/main/kotlin/edu/artic/db/RetrofitAppDataServiceProvider.kt
Cliabhach: By "[replace] with AppDataManager class" you're suggesting I inject `AppDataManager` then subscribe to the `AppDataManager.dataObject`, correct? That seems workable.

agrosner: possibly, but it'll lead to circular dependencies, which is not allowed since `AppDataManager` already depends on this class.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
agrosner: right, just a warning. any chain you expect to continue or depend on this object will not emit a value _until_ something is put into the DB, which should be fine.

db/src/main/kotlin/edu/artic/db/RetrofitAppDataServiceProvider.kt
Cliabhach: What if I move our cannonical `dataObject` field into the DAO class? That would solve the dependency concern, but at a cost.

Cliabhach:
I'm going to remove the changes involving the `dataObject` field. We can solve that separately.

PR #84, opened 2018-09-04
Audio lookup: correct current audio display (AIC-327, AIC-329)

Cliabhach:


sam33rdhakal:
@Cliabhach It would be easy to review the work if you leave the summary of the work in comments (instead of JIRA ticket numbers). Since it is an open source project, I recommend not to use JIRA ticket numbers.

Cliabhach:
Sorry, I was a bit rushed when I made the PR and skipped the description. I'll add that now.

In the matter of Jira tickets, it's hardly atypical to use private issue numbers in public projects. The work needs to be correlated with whichever bug report or feature request induced it, and this offers a solid mechanism for verifying that.

sam33rdhakal:
JIRA ticket number in the description should be fine as long as merge comments describe the changes.

audio/src/main/java/edu/artic/audio/AudioLookupFragment.kt
```diff
@@ -73,25 +74,31 @@ class AudioLookupFragment : BaseViewModelFragment<AudioLookupViewModel>() {
...
                 .disposedBy(disposeBag)
 
     }
 
-    private fun sendToPlayerFragment(foundAudio: LookupResult.FoundAudio) {
+    private fun playAndDisplay(foundAudio: LookupResult.FoundAudio) {
         audioService.subscribeBy {
             it?.playPlayer(foundAudio.hostObject)
+            val host = baseActivity
+            host.runOnUiThread {
```
agrosner: we should just do an `observeOn(AndroidSchedulers.main())` on `audioService` subject.

Cliabhach: Won't that make the `playPlayer` call execute on the main thread too?

sam33rdhakal: Resolved it since @Cliabhach updated the code.

PR #90, opened 2018-09-05
Leave tour

sam33rdhakal:
Updates:
* Displays leave tour confirmation dialog
* Ends the tour display mode when a user confirms to leave the tour

### DIFF
--- a/map/src/main/kotlin/edu/artic/map/MapFragment.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapFragment.kt
@@ -72,8 +72,11 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
     private lateinit var buildingGroundOverlay: GroundOverlay
     private var groundOverlayGenerated: Subject<Boolean> = BehaviorSubject.createDefault(false)
     private var mapClicks: Subject<Boolean> = PublishSubject.create()
-
-    private val tour: ArticTour? get() = requireActivity().intent?.extras?.getParcelable(MapActivity.ARG_TOUR)
+    private var tour: ArticTour?
+        get() = requireActivity().intent?.extras?.getParcelable(MapActivity.ARG_TOUR)
+        set(value) {
+            requireActivity().intent?.putExtra(MapActivity.ARG_TOUR, value)
+        }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
--- a/map/src/main/kotlin/edu/artic/map/MapFragment.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapFragment.kt
@@ -4,24 +4,15 @@ import android.content.Context
 import android.graphics.Bitmap
 import android.os.Bundle
 import android.support.annotation.AnyThread
+import android.support.annotation.UiThread
 import android.support.v4.app.Fragment
+import android.support.v7.app.AlertDialog
 import android.view.View
-import com.fuzz.rx.defaultThrottle
-import com.fuzz.rx.disposedBy
-import com.fuzz.rx.filterFlatMap
-import com.fuzz.rx.filterTo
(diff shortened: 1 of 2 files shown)


map/src/main/kotlin/edu/artic/map/MapViewModel.kt
```diff
@@ -38,7 +38,7 @@ class MapViewModel @Inject constructor(val mapMarkerConstructor: MapMarkerConstr
     private val floor: Subject<Int> = BehaviorSubject.createDefault(1)
     private val focus: Subject<MapFocus> = BehaviorSubject.create()
 
-    val displayMode: Subject<MapDisplayMode> = BehaviorSubject.create()
+    val displayMode: Subject<MapDisplayMode> = PublishSubject.create()
```
Cliabhach: This prevents caching of the display mode, right? Good idea.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -72,8 +64,12 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
     private lateinit var buildingGroundOverlay: GroundOverlay
     private var groundOverlayGenerated: Subject<Boolean> = BehaviorSubject.createDefault(false)
     private var mapClicks: Subject<Boolean> = PublishSubject.create()
-
-    private val tour: ArticTour? get() = requireActivity().intent?.extras?.getParcelable(MapActivity.ARG_TOUR)
+    private var leaveTourDialog: AlertDialog? = null
+    private var tour: ArticTour?
+        get() = requireActivity().intent?.extras?.getParcelable(MapActivity.ARG_TOUR)
+        set(value) {
+            requireActivity().intent?.extras?.putParcelable(MapActivity.ARG_TOUR, value)
```
Cliabhach: Please switch to `intent?.putExtra` - that properly handles the null check for the `extras` property. See https://github.com/aosp-mirror/platform_frameworks_base/blob/1c77a11db2/core/java/android/content/Intent.java#L8575

map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -72,8 +64,12 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
     private lateinit var buildingGroundOverlay: GroundOverlay
     private var groundOverlayGenerated: Subject<Boolean> = BehaviorSubject.createDefault(false)
     private var mapClicks: Subject<Boolean> = PublishSubject.create()
-
-    private val tour: ArticTour? get() = requireActivity().intent?.extras?.getParcelable(MapActivity.ARG_TOUR)
+    private var leaveTourDialog: AlertDialog? = null
```
Cliabhach: Why not use a `DialogFragment`?

Cliabhach: If it's going to be an `AlertDialog`, you should probably call `leaveTourDialog?.dismiss()` in `onDestroy` or `onDestroyView`, just in case the Activity has to be recreated or killed before a dialog button is pressed.

sam33rdhakal: Added `leaveTourDialog?.dismiss()` in `onDestroyView`.
Curious why you suggested `DialogFragment`?

map/src/main/kotlin/edu/artic/map/MapFragment.kt
sam33rdhakal: Now uses `intent?.putExtra`.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
Cliabhach: Just to keep it in line with the activity lifecycle. Your change covers that, so it's fine.

PR #94, opened 2018-09-05
Basic Search backend (AIC-342)

Cliabhach:
This adds a new `SearchApi` class, along with a `SearchServiceProvider` abstracting that away from the ViewModel.

I've left off the actual execution of the api calls and the binding of search results for now; I expect those'll need more attention so I excluded them from this PR.

### DIFF
--- a/search/src/main/java/artic/edu/search/SearchFragment.kt
+++ b/search/src/main/java/artic/edu/search/SearchFragment.kt
@@ -1,7 +1,11 @@
 package artic.edu.search
 
+import com.fuzz.rx.bindTo
+import com.fuzz.rx.disposedBy
+import com.jakewharton.rxbinding2.widget.textChanges
 import edu.artic.analytics.ScreenCategoryName
 import edu.artic.viewmodel.BaseViewModelFragment
+import kotlinx.android.synthetic.main.search_app_bar_layout.view.*
 import kotlin.reflect.KClass
 
 class SearchFragment : BaseViewModelFragment<SearchViewModel>() {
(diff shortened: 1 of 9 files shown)


search/src/main/java/artic/edu/search/SearchModule.kt
```diff
@@ -39,4 +46,34 @@ abstract class SearchModule {
...
+
+    @Module
+    companion object {
+
+        @JvmStatic
+        @Provides
+        @Singleton
+        @Named(SEARCH_CLIENT_API)
+        fun provideClient(): OkHttpClient {
+            return OkHttpClient.Builder().build()
```
agrosner: any reason we don't reuse existing client? also if we're using retrofit from `ApiModule`, this is effectively ignored.

Cliabhach: Oh, that just slipped through from the first draft of the commit. I wanted to create a copy of the retrofit and set the dataObject url on that, but @raigex suggested this particular type of reuse.

It can be deleted safely.

search/src/main/java/artic/edu/search/SearchFragment.kt
```diff
@@ -19,4 +23,17 @@ class SearchFragment : BaseViewModelFragment<SearchViewModel>() {
 
     override val customToolbarColorResource: Int
         get() = R.color.greyText
+
+    override fun setupBindings(viewModel: SearchViewModel) {
+        super.setupBindings(viewModel)
+
+        toolbar!!.searchEditText
```
raigex: should probably use toolbar?.let OR use the toolbar from kotlinx

Cliabhach: I couldn't get the kotlinx one to import, didn't want to diagnose. I'll leave it for now.

PR #95, opened 2018-09-06
Basic UI work for search

raigex:
Basically boilerplate code to be used for search.


search/src/main/java/artic/edu/search/SearchResultsManager.kt
```diff
@@ -0,0 +1,61 @@
...
+                    ArticSearchResult(
+                            it,
+                            emptyList(),
+                            emptyList(),
+                            emptyList()
+                    )
+                }.bindTo(currentSearchResults)
+    }
+
+    private fun setupExmptyTextSearchFlow() {
```
Cliabhach: Rename `Exmpty` -> `Empty`

PR #97, opened 2018-09-06
Link from audio to search (AIC-328)

Cliabhach:
Follow-up to #89, ensuring that the search icon links to the search screen.


audio/src/main/java/edu/artic/audio/AudioLookupFragment.kt
```diff
@@ -94,40 +98,42 @@ class AudioLookupFragment : BaseViewModelFragment<AudioLookupViewModel>() {
         registerNumPadSubscription()
 
         getAudioServiceObservable(fm = childFragmentManager)
-                .bindTo(audioService)
+                .bindTo(viewModel.audioService)
```
raigex: Maybe we should find another way to do this. Idea is a field `var audioService : AudioService ? = null` that you set and unset. might need more checks but should be more robust

PR #100, opened 2018-09-07
Reduce duplicate RX subscriptions (AIC-349)

Cliabhach:
This is intended to resolve some visual problems and some (at first glance inconsistent) behavior.

Sample issue: navigate in the following pattern:

1. Hit 'audio'
2. Hit 'home' (do NOT use the back button)
3. Hit 'audio'
3a. Each button press on the number pad will be registered twice.
3b. This includes the 'Go' button - we can go only once, and the second invocation triggers a crash

Each subsequent load of the 'audio' screen in the same manner as steps 1. and 2. will register an extra subscription and cause more trouble. The number of duplicate button presses is capped at 5 but the subscription count is unlimited.

Work done in this branch:

*    Document navigationDisposeBag in the code itself
*    Restrict standard binding to one resume per create-destroy lifecycle
*    Run `Window.setStatusBarColor` earlier than currently-done (c.f. comments added in 9c408030411d3e7646b73d0)
*    Prevent duplicate launches of `AudioActivity` that were causing duplicate fragments to register duplicate on-click observers and duplicate on-navigate observers

### DIFF
--- a/viewmodel/src/main/kotlin/edu/artic/viewmodel/BaseViewModelFragment.kt
+++ b/viewmodel/src/main/kotlin/edu/artic/viewmodel/BaseViewModelFragment.kt
@@ -102,5 +102,10 @@ abstract class BaseViewModelFragment<TViewModel : BaseViewModel> : BaseFragment(
 
     protected open fun setupBindings(viewModel: TViewModel) = Unit
 
+    /**
+     * # Remember to dispose your nav bindings with [navigationDisposeBag], not [disposeBag].
+     *
+     * @see NavViewViewModel
+     */
     protected open fun setupNavigationBindings(viewModel: TViewModel) = Unit
 }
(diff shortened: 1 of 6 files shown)


viewmodel/src/main/kotlin/edu/artic/viewmodel/BaseViewModelFragment.kt
```diff
@@ -64,6 +64,8 @@ abstract class BaseViewModelFragment<TViewModel : BaseViewModel> : BaseFragment(
     override fun onResume() {
         super.onResume()
         if (isViewJustCreated) {
+            // Without this assignment, the screen will flicker (whether it's noticeable depends on the subclass)
+            isViewJustCreated = false
```
raigex: Cant believe I missed this

PR #101, opened 2018-09-07
Ignore duplicate clicks on bottom nav (AIC-363)

Cliabhach:
This is a followup to #100. I noticed that we go through a wasteful `pause`-`resume` cycle whenever the icon for the current section of the app is tapped.

In addition, such intents overwrite the original launch intent and unnecessarily encumber diagnosis of navigation graph issues. This PR includes the logic needed to `preventReselection()` - see inline documentation for more details.


Cliabhach:
@raigex Good question, we should probably raise it with the https://github.com/art-institute-of-chicago/aic-mobile-ios maintainers on Monday.

PR #103, opened 2018-09-07
[Feature] Loads the searched object in the map

sam33rdhakal:
This work basically includes loading the default suggested map items to map.

### DIFF
--- a/map/src/main/kotlin/edu/artic/map/MapFragment.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapFragment.kt
@@ -7,7 +7,6 @@ import android.support.annotation.AnyThread
 import android.support.annotation.UiThread
 import android.support.v4.app.Fragment
 import android.support.v7.app.AlertDialog
-import android.util.Log
 import android.view.View
 import com.fuzz.rx.*
 import com.google.android.gms.maps.CameraUpdateFactory
@@ -322,7 +321,7 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
 
         viewModel.selectedArticObject
                 .withLatestFrom(viewModel.displayMode) { selected, mapMode -> selected to mapMode }
-                .filterFlatMap({ (_, mapMode) -> mapMode !is MapDisplayMode.Tour }, { (selected) -> selected })
+                .filterFlatMap({ (_, mapMode) -> mapMode is MapDisplayMode.CurrentFloor }, { (selected) -> selected })
                 .subscribeBy { selected ->
                     displayFragmentInInfoContainer(MapObjectDetailsFragment.create(selected))
                 }
@@ -384,7 +383,7 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
      * Removes the fragment information below the map, and readjusts the padding back to normal.
      */
     @UiThread
-    private fun hideFragmentInInfoContainer(tag : String = OBJECT_DETAILS) {
+    private fun hideFragmentInInfoContainer(tag: String = OBJECT_DETAILS) {
         val supportFragmentManager = requireActivity().supportFragmentManager
         supportFragmentManager
                 .findFragmentByTag(tag)
--- a/map/src/main/kotlin/edu/artic/map/MapFragment.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapFragment.kt
@@ -74,11 +74,6 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
             requireActivity().intent?.putExtra(MapActivity.ARG_SEARCH_OBJECT, value)
         }
 
-    private var searchAmenityType: String?
-        get() = requireActivity().intent?.extras?.getString(MapActivity.ARG_SEARCH_AMENITY_TYPE)
-        set(value) {
-            requireActivity().intent?.putExtra(MapActivity.ARG_SEARCH_AMENITY_TYPE, value)
-        }
 
     private var tour: ArticTour?
         get() = requireActivity().intent?.extras?.getParcelable(MapActivity.ARG_TOUR)
@@ -425,6 +420,7 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
                     .setNegativeButton(getString(R.string.leave)) { dialog, _ ->
                         tour = null
                         viewModel.leaveTour()
+                        viewModel.loadMapDisplayMode(tour, startTourStop)
                         dialog.dismiss()
                         audioService.subscribe {
                             it.stopPlayer()
@@ -444,7 +440,10 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
     override fun onResume() {
         super.onResume()
         mapView.onResume()
-        viewModel.loadMapDisplayMode(tour, startTourStop, searchObject, searchAmenityType)
+        searchObject?.let {
+            viewModel.searchManager.selectedObject.onNext(Optional(it))
+        }
+        viewModel.loadMapDisplayMode(tour, startTourStop)
     }
 
     override fun onPause() {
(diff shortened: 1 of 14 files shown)


map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -66,6 +66,14 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
     private var groundOverlayGenerated: Subject<Boolean> = BehaviorSubject.createDefault(false)
     private var mapClicks: Subject<Boolean> = PublishSubject.create()
     private var leaveTourDialog: AlertDialog? = null
+
+    private var searchObject: ArticObject?
```
Cliabhach: What do you think of moving this into `SearchManager`?

map/src/main/kotlin/edu/artic/map/SearchObjectDetailsFragment.kt
```diff
@@ -0,0 +1,164 @@
...
+
+        viewModel.title
+                .bindToMain(searchObjectTitle.text())
+                .disposedBy(disposeBag)
+
+        viewModel.galleryLocation
+                .bindToMain(subtitle.text())
+                .disposedBy(disposeBag)
+
+        viewModel.image.subscribe {
```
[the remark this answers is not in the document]
Cliabhach: We should be subscribing explicitly on the main thread, so Glide doesn't crash.

map/src/main/kotlin/edu/artic/map/SearchObjectDetailsFragment.kt
```diff
@@ -0,0 +1,164 @@
...
+import io.reactivex.android.schedulers.AndroidSchedulers
+import io.reactivex.rxkotlin.subscribeBy
+import io.reactivex.rxkotlin.withLatestFrom
+import io.reactivex.subjects.BehaviorSubject
+import io.reactivex.subjects.Subject
+import kotlinx.android.synthetic.main.fragment_search_object_details.*
+import kotlin.reflect.KClass
+
+/**
+ * Displays the object details of the searched item.
```
Cliabhach: We really ought to have some disambiguation with `MapObjectDetailsFragment`.

map/src/main/kotlin/edu/artic/map/SearchObjectDetailsFragment.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    private fun displayPlayButton() {
+        playCurrent.visibility = View.VISIBLE
+        pauseCurrent.visibility = View.INVISIBLE
+    }
+
+
+    companion object {
+        private val ARG_MAP_OBJECT = SearchObjectDetailsFragment::class.java.simpleName
```
Cliabhach: ....`ARG_MAP_OBJECT`?

map/src/main/kotlin/edu/artic/map/SearchObjectDetailsViewModel.kt
```diff
@@ -0,0 +1,155 @@
...
+import edu.artic.media.audio.AudioPlayerService
+import edu.artic.media.audio.preferredLanguage
+import edu.artic.viewmodel.BaseViewModel
+import io.reactivex.rxkotlin.withLatestFrom
+import io.reactivex.subjects.BehaviorSubject
+import io.reactivex.subjects.Subject
+
+
+/**
+ * @author Sameer Dhakal (Fuzz)
```
Cliabhach: Add a TODO linking this to the other file....

map/src/main/kotlin/edu/artic/map/rendering/MapItemRenderers.kt
```diff
@@ -81,7 +88,12 @@ class SpacesMapItemRenderer(articMapAnnotationDao: ArticMapAnnotationDao)
     private val textMarkerGenerator: TextMarkerGenerator  by lazy { TextMarkerGenerator(context) }
 
     override fun getItems(floor: Int, displayMode: MapDisplayMode): Flowable<List<ArticMapAnnotation>> {
-        return articMapAnnotationDao.getTextAnnotationByTypeAndFloor(ArticMapTextType.SPACE, floor = floor.toString()) // TODO: switch to int
+        return when(displayMode) {
+            is MapDisplayMode.Search<*> ->
+                listOf<ArticMapAnnotation>().asFlowable()
```
Cliabhach: There seems to be a lot of these `listOf<T>().asFlowable()` in this form. Perhaps we could pull those out into a common function, or maybe even a constant object. @agrosner, what do you think?

map/src/main/res/layout/fragment_search_object_details.xml
```diff
@@ -0,0 +1,133 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Cliabhach: This file will need to act more like a carousel in future, right? If not, you should really extract some common styles and compact down this file.

search/src/main/java/artic/edu/search/DefaultSearchSuggestionsFragment.kt
```diff
@@ -69,10 +71,11 @@ class DefaultSearchSuggestionsFragment : BaseViewModelFragment<DefaultSearchSugg
                     when (navigation.endpoint) {
                         is DefaultSearchSuggestionsViewModel.NavigationEndpoint.ArticObjectDetails -> {
                             val o = (navigation.endpoint as DefaultSearchSuggestionsViewModel.NavigationEndpoint.ArticObjectDetails).articObject
-                            Navigation.findNavController(requireActivity(), R.id.container).navigate(
-                                    R.id.goToSearchAudioDetails,
-                                    SearchAudioDetailFragment.argsBundle(o)
-                            )
+                            val mapIntent = NavigationConstants.MAP.asDeepLinkIntent().apply {
+                                putExtra("ARG_SEARCH_OBJECT", o)
```
Cliabhach: .....Wonder if that string shouldn't be in `NavigationConstants` too.

map/src/main/kotlin/edu/artic/map/SearchObjectDetailsFragment.kt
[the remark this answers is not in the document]
sam33rdhakal: Good catch. Updated.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
sam33rdhakal: `SearchManager` already has a subject that persists the `searchObject`. This field reads the object from the activity's bundle and passes it to `SearchManager` if the value is not null.

map/src/main/kotlin/edu/artic/map/SearchObjectDetailsFragment.kt
sam33rdhakal: Are you talking about the name of the fragment or the functionality? Once I integrate the map annotation it will have more features (e.g. cafe carousel etc.) than the `MapObjectDetailsFragment`.

map/src/main/kotlin/edu/artic/map/SearchObjectDetailsFragment.kt
sam33rdhakal: Renamed.

map/src/main/res/layout/fragment_search_object_details.xml
sam33rdhakal: You are right. This fragment will have carousel logic in the future.

search/src/main/java/artic/edu/search/DefaultSearchSuggestionsFragment.kt
sam33rdhakal: That sounds right. Moved to `NavigationConstants`

map/src/main/kotlin/edu/artic/map/MapFragment.kt
Cliabhach: Then this should really be a function. How about `getLatestSearchObject()`?

PR #104, opened 2018-09-07
Search Results

raigex:


raigex:
Circle CI fails cuz of Constraint Lint check
UI looks right will look into "constraints" at a later point

search/src/main/java/artic/edu/search/SearchResultsSuggestedViewModel.kt
```diff
@@ -2,23 +2,104 @@ package artic.edu.search
...
+
+        Observables
+                .combineLatest(
+                        dynamicCells,
+                        Observable.just(listOf(
+                                SearchResultAmenitiesCellViewModel(R.drawable.ic_icon_amenity_map_restaurant),
+                                SearchResultAmenitiesCellViewModel(R.drawable.ic_icon_amenity_map_lounge),
+                                SearchResultAmenitiesCellViewModel(R.drawable.ic_icon_amenity_map_shop),
+                                SearchResultAmenitiesCellViewModel(R.drawable.ic_icon_amenity_map_restroom),
+                                SearchResultAmenitiesCellViewModel(0))
```
sam33rdhakal: I couldn't understand why we need the `SearchResultAmenitiesCellViewModel` without the icon here? Is this for breaking the line for amenities icon? 

I did exactly the same work in `DefaultSearchSuggestionsFragment` for displaying on the map suggestions. Since the UI for **on the map** section for default search suggestions and search suggestions is exactly the same,I think we should share the viewModels.

search/src/main/java/artic/edu/search/SearchResultsAdapter.kt
```diff
@@ -1,33 +1,92 @@
...
                         .disposedBy(item.viewDisposeBag)
             }
+            is SearchResultCircularCellViewModel -> {
+                item.imageUrl
+                        .subscribe {
+                            Glide.with(this)
+                                    .load(it)
+                                    .apply(RequestOptions.circleCropTransform())
+                                    .into(circularImage)
+                        }
```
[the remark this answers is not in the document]
raigex: Not really as the Glide.into auto works on main thread. especially since it's an image view

search/src/main/java/artic/edu/search/SearchResultsSuggestedViewModel.kt
raigex: We def should, I thought I only needed to copy small parts of it. but it looks like it was much larger. Hope to do that at a later point during a refactor

PR #111, opened 2018-09-12
Minor splash improvements (AIC-380)

Cliabhach:
I know that we've got a much nicer splash screen planned out. Nevertheless, I felt there were some minor improvements we could make now without significant churn. Additionally, we will need the load-level error handling logic no matter what kind of splash screen we end up displaying.

* Progress percentage is fixed-width
* Errors thrown during load are shown in place of progress text and do not crash app
* Progress text is larger font size and black instead of grey for heightened contrast

NB: only the first error will appear on screen. I feel this is acceptable for now.

### DIFF
--- a/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
+++ b/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
@@ -216,19 +216,36 @@ class AppDataManager @Inject constructor(
      * [primary stuff][getBlob]. As a side-effect, the injected [serviceProvider]
      * (and not `this` manager) is responsible for determining precisely
      * how many exhibitions and events we parse when this method is called.
+     *
+     * Observers on this should implement `onError` to handle I/O errors. We
+     * use [Observable.onErrorReturn] here to make sure that even if multiple
+     * errors occur at a low-level, only one call would go out to that `onError`.
      */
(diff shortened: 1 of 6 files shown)


db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -216,19 +216,36 @@ class AppDataManager @Inject constructor(
      * [primary stuff][getBlob]. As a side-effect, the injected [serviceProvider]
      * (and not `this` manager) is responsible for determining precisely
      * how many exhibitions and events we parse when this method is called.
+     *
+     * Observers on this should implement `onError` to handle I/O errors. We
+     * use [Observable.onErrorReturn] here to make sure that even if multiple
+     * errors occur at a low-level, only one call would go out to that `onError`.
      */
     private fun loadSecondaryData(): Observable<Int> {
         return Observable.zip(
```
raigex: This can be changed to Observables.zip

Cliabhach: I think that _might_ be out of scope. We'll come back for that when we need to change the loading screen animation (part of AIC-17).

Cliabhach:
Ah, I _knew_ there was something. I'll merge it in now.

PR #123, opened 2018-09-14
Update testing field

raigex:


Cliabhach:
Anyway, please wait until we've got the ok from Circle CI

PR #124, opened 2018-09-14
Rename search package

Cliabhach:
Unlike the other packages, this was placed under the name

`artic.edu.search`

instead of

`edu.artic.search`.

This PR corrects the oversight.


Cliabhach:
Retrying because CircleCI can't handle dependency download timeouts with our current config.

PR #125, opened 2018-09-17
AIC-75 Bugfixes

raigex:


search/src/main/java/edu/artic/search/SearchAudioDetailViewModel.kt
```diff
@@ -58,7 +67,7 @@ class SearchAudioDetailViewModel @Inject constructor(private val analyticsTracke
 
         articObjectObservable
                 .map { it.location.orEmpty() }
-                .map { it.isEmpty() }
+                .map { !it.isEmpty() }
```
Cliabhach: We could use `it.isNotEmpty()` too, right?

search/src/main/res/layout/fragment_search_audio_detail.xml
```diff
@@ -54,6 +54,7 @@
                     android:contentDescription="@null"
                     android:fitsSystemWindows="true"
                     android:src="@color/audioBackground"
+                    android:adjustViewBounds="false"
```
Cliabhach: This was the default state, I think.

search/src/main/java/edu/artic/search/SearchAudioDetailViewModel.kt
```diff
@@ -76,6 +85,13 @@ class SearchAudioDetailViewModel @Inject constructor(private val analyticsTracke
                 .bindTo(playAudioVisible)
                 .disposedBy(disposeBag)
 
+        articObjectObservable
+                .filterFlatMap({it.gallery != null} , {it.gallery!!})
+                .filter { it.number != null }
+                .map { "Gallery ${it.number.toString()}" } // TODO: use localizer
```
Cliabhach: We can at least use a string resource here.

search/src/main/java/edu/artic/search/SearchAudioDetailViewModel.kt
raigex: true. updated

search/src/main/res/layout/fragment_search_audio_detail.xml
raigex: most likely

search/src/main/java/edu/artic/search/SearchAudioDetailViewModel.kt
raigex: updated to pull generation of this full field into Fragment under protest

PR #126, opened 2018-09-17
Refactors search to map flow (AIC-163)

sam33rdhakal:
Updates:
* Refactors the way objects are sent to the map. `SearchObjectDetailsFragment` now only accepts the proxy object `ArticSearchArtworkObject`. 
* Hides the playback control from search details if no audio translation is available

### DIFF
--- a/search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
+++ b/search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
@@ -93,20 +93,21 @@ open class SearchBaseViewModel @Inject constructor(
                 navigateTo.onNext(Navigate.Forward(NavigationEndpoint.Web))
             }
             is SearchCircularCellViewModel -> {
-                viewModel.artWork?.let { arcticObject ->
-                    /* Convert the ArcticObject to ArcticSearchArtworkObject*/
-                    val arcticSearchArtworkObject = ArticSearchArtworkObject(
-                            artworkId = arcticObject.id.toString(),
-                            audioObject = arcticObject,
-                            title = arcticObject.title,
-                            thumbnailUrl = arcticObject.thumbUrl,
-                            imageUrl = arcticObject.largeImageUrl,
-                            artistDisplay = arcticObject.artistCulturePlaceDelim,
-                            location = arcticObject.location,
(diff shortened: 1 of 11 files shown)


search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
```diff
@@ -94,13 +93,20 @@ open class SearchBaseViewModel @Inject constructor(
                 navigateTo.onNext(Navigate.Forward(NavigationEndpoint.Web))
             }
             is SearchCircularCellViewModel -> {
-                viewModel.artWork?.let { articObject ->
-                    navigateTo.onNext(
-                            Navigate.Forward(
-                                    NavigationEndpoint.ArtworkOnMap(articObject)
-                            )
-                    )
+                viewModel.artWork?.let { arcticObject ->
```
Cliabhach: Please replace all instances of `arctic` with `artic`.

search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
```diff
@@ -94,13 +93,20 @@ open class SearchBaseViewModel @Inject constructor(
...
+                    val arcticSearchArtworkObject = ArticSearchArtworkObject(
+                            artworkId = arcticObject.id.toString(),
+                            audioObject = arcticObject,
+                            title = arcticObject.title,
+                            thumbnailUrl = arcticObject.thumbUrl,
+                            imageUrl = arcticObject.largeImageUrl,
+                            artistDisplay = arcticObject.artistCulturePlaceDelim,
+                            location = arcticObject.location,
+                            floor = arcticObject.floor,
+                            gallery = null)
```
Cliabhach: Put close parenthesis `)` on its own line.

Cliabhach: add a comment too saying that we don't pass 'gallery' because we don't need it at this time, or something like that.

search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
sam33rdhakal: Updated.

PR #129, opened 2018-09-17
Make the 'see all' button in search results actually do something (AIC-440)

raigex:


search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
```diff
@@ -123,4 +123,8 @@ open class SearchBaseViewModel @Inject constructor(
             }
         }
     }
+
+    open fun onClickSeeAll(header: Header) {
```
Cliabhach: I think this _could_'ve been an interface or abstract function, but I'm not going to be picky.

PR #131, opened 2018-09-18
Fade out old pins on map change (AIC-288)

raigex:


Cliabhach:
I suppose the real question is "does it look good". I'll try out a build first - I don't think that will take more than a couple minutes.

PR #139, opened 2018-09-19
App bar subtitle [AIC-63]

sam33rdhakal:
Adds subtitle to app bar layout.

### DIFF
--- a/ui/src/main/kotlin/edu/artic/view/ArticMainAppBarLayout.kt
+++ b/ui/src/main/kotlin/edu/artic/view/ArticMainAppBarLayout.kt
@@ -9,8 +9,6 @@ import com.fuzz.rx.DisposeBag
 import com.fuzz.rx.defaultThrottle
 import com.fuzz.rx.disposedBy
 import com.jakewharton.rxbinding2.view.clicks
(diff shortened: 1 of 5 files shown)


ui/src/main/kotlin/edu/artic/view/ArticMainAppBarLayout.kt
```diff
@@ -20,8 +18,8 @@ import kotlinx.android.synthetic.main.view_app_bar_layout.view.*
  * this app.
  */
 class ArticMainAppBarLayout(context: Context, attrs: AttributeSet? = null) : AppBarLayout(context, attrs) {
-    private val disposeBag : DisposeBag = DisposeBag()
-    private var clickConsumer : Consumer<Unit>? = null
+    private val disposeBag: DisposeBag = DisposeBag()
```
raigex: where is the disposebag clear? I know I created it, but can you add disposebag.clear remove from window?

sam33rdhakal: It's inside `onDetachedFromWindow`.

PR #140, opened 2018-09-19
Map Location Prompt (AIC-77)

raigex:
Shows a prompt unless location permission has been requested

### DIFF
--- a/location/src/main/res/values/strings.xml
+++ b/location/src/main/res/values/strings.xml
@@ -7,4 +7,6 @@
     <string name="locationSettingsLocationNeverAsked">Location Settings</string>
     <string name="locationSettingsLocationAllowed">Location Allowed</string>
     <string name="locationSettingsLocationDenied">Location Denied</string>
+    <string name="locationPromptNotNow">Now Right Now</string>
+    <string name="locationPromptOk">OK, Use My Location</string>
 </resources>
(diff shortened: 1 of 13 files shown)


map/src/main/kotlin/edu/artic/map/MapViewModel.kt
```diff
@@ -144,6 +152,17 @@ class MapViewModel @Inject constructor(val mapMarkerConstructor: MapMarkerConstr
     private val lockVisibleRegion: Subject<Boolean> = BehaviorSubject.createDefault(false)
 
     init {
+        locationService.hasRequestedPermissionAlready
```
sam33rdhakal: Looks like it is to debug code. Please remove it.

location/src/main/res/values/strings.xml
```diff
@@ -7,4 +7,6 @@
     <string name="locationSettingsLocationNeverAsked">Location Settings</string>
     <string name="locationSettingsLocationAllowed">Location Allowed</string>
     <string name="locationSettingsLocationDenied">Location Denied</string>
+    <string name="locationPromptNotNow">Now Right Now</string>
```
Cliabhach: We should change this to `Not Right Now`. I can do that when I'm adding the Spanish and Chinese translations for AIC-431 or similar.

PR #149, opened 2018-09-20
Removes tour object from bundle once used (AIC-452)

sam33rdhakal:
Changes :
* Clears the tour object from the bundle.

Since tour object was not removed, the map was presented in tour mode every time the user returned to map (it occurred only when `MapFragment` was resumed).

### DIFF
--- a/map/src/main/kotlin/edu/artic/map/MapFragment.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapFragment.kt
@@ -71,26 +71,30 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
     private var leaveTourDialog: LeaveCurrentTourDialogFragment? = null
 
     private fun getLatestSearchObject(): ArticSearchArtworkObject? {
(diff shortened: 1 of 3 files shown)


map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -82,14 +83,14 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
     }
 
     private fun getLatestTourObject(): ArticTour? {
-        val tour: ArticTour? = requireActivity().intent?.getParcelableExtra(MapActivity.ARG_TOUR)
-        requireActivity().intent?.extras?.remove(MapActivity.ARG_TOUR)
+        val tour: ArticTour? = requireActivity().intent?.getParcelableExtra(NavigationConstants.ARG_TOUR)
+        requireActivity().intent?.removeExtra(NavigationConstants.ARG_TOUR)
```
Cliabhach: What do you think of using a local variable for `requireActivity().intent?`?

PR #153, opened 2018-09-21
Fade in markers on floor change (AIC-288)

Cliabhach:
I think the timing of the animations may need a little work. For now, I've set a duration of 500 milliseconds even though the Swift code (in [MapView.swift](https://github.com/art-institute-of-chicago/aic-mobile-ios/blob/0b08e746ca92e/aic/aic/ViewControllers%2BViews/Sections/Map/MapView.swift)) uses 250 milliseconds. More critical though is the actual code changes:

1. Create and use a new `Marker::fadeIn` function
2. Deduplication of visible region changes (prevents stale region info from sticking around)
3. Properly adjust item coords before checking if they're on-screen
4. Small yet handy updates to documentation

Item 3. gave me the most grief - it's the code change in `ObjectsMapItemRenderer.kt` line 108. This logic is irrelevant when there's a constant stream of updates, like when the map's being dragged around. But if you switch floors it is absolutely crucial to detecting which objects need to be drawn with full markers and which should be little dots.

### DIFF
--- a/map/src/main/kotlin/edu/artic/map/GoogleMapUtil.kt
+++ b/map/src/main/kotlin/edu/artic/map/GoogleMapUtil.kt
@@ -14,6 +14,7 @@ import com.google.android.gms.maps.model.LatLngBounds
 import com.google.android.gms.maps.model.Marker
 import edu.artic.base.utils.statusBarHeight
 import edu.artic.db.models.ArticObject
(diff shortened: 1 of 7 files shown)


Cliabhach:
call that timing - i think it just rejected the review. Thanks @raigex!

PR #135, opened 2018-09-18
Access member card

sam33rdhakal:
Updates
* Adds access member card feature 
* User will be now able to access the member card by entering the memberID and zip code

PR #156, opened 2018-09-21
Handle missing `Member Info API` gracefully

Cliabhach:
This adds a bunch of documentation to the member api and offers friendly error messages if the build was set up incorrectly. Follow-up to #135 and #147 .

### DIFF
--- a/info/src/main/java/edu/artic/info/InfoModule.kt
+++ b/info/src/main/java/edu/artic/info/InfoModule.kt
@@ -77,20 +77,32 @@ abstract class InfoModule {
         fun provideRetrofit(@Named(ApiModule.BLOB_CLIENT_API)
                             client: OkHttpClient,
                             executor: Executor)
-                : Retrofit = constructRetrofit(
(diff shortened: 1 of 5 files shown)


info/src/main/java/edu/artic/info/InfoModule.kt
```diff
@@ -77,20 +77,32 @@ abstract class InfoModule {
...
-                baseUrl = BuildConfig.MEMBER_INFO_API,
-                client = client,
-                executor = executor,
-                customConfiguration = {
-                    addConverterFactory(SimpleXmlConverterFactory.createNonStrict(
-                            Persister(AnnotationStrategy())
-                    ))
-                })
+                : Retrofit? {
+            return if (BuildConfig.MEMBER_INFO_API.indexOf("https") < 0) {
```
sam33rdhakal: This works fine but using `contains` would make this more readable IMO.

Cliabhach: Oh, good point. Let me do that.

PR #159, opened 2018-09-24
Prevent usage of Zoom Level 6 (AIC-466)

raigex:


map/src/main/kotlin/edu/artic/map/MapConstructs.kt
```diff
@@ -16,7 +16,7 @@ const val ZOOM_LANDMARK: ZoomLevel = 17.5f
 const val ZOOM_DEPARTMENTS: ZoomLevel = 18.0f
 const val ZOOM_DEPARTMENT_AND_SPACES: ZoomLevel = 19.0f
 const val ZOOM_INDIVIDUAL: ZoomLevel = 21.0f
```
Cliabhach: Shouldn't this go down too?

PR #160, opened 2018-09-24
[Url] replace utm_source iphone with android (AIC-386)

raigex:


db/src/main/kotlin/edu/artic/db/models/ArticDataObject.kt
```diff
@@ -24,5 +24,28 @@ data class ArticDataObject(
         @Json(name = "membership_url") val membershipUrl: String?,
         @Json(name = "website_url") val websiteUrl: String?,
         @Json(name = "tickets_url") val ticketsUrl: String?,
-        @PrimaryKey val id : Int = 0
-)
\ No newline at end of file
+        @PrimaryKey val id: Int = 0
+) {
+
+    val memberShipUrlAndroid: String
```
Cliabhach: `memberShip` -> `membership`

info/src/main/java/edu/artic/info/MuseumInformationViewModel.kt
```diff
@@ -46,7 +46,7 @@ class MuseumInformationViewModel @Inject constructor(
 
         dataObjectDao.getDataObject()
                 .toObservable()
-                .filterFlatMap({ it.ticketsUrl != null }, { it.ticketsUrl!! })
+                .map { it.ticketsUrlAndroid }
```
Cliabhach: Don't we have to filter out empty strings here?

raigex: yes we do done.

db/src/main/kotlin/edu/artic/db/models/ArticDataObject.kt
raigex: done

PR #162, opened 2018-09-24
[Map] Remove excess padding from geo-coordinates

raigex:


map/src/main/kotlin/edu/artic/map/rendering/MapProjectionHelper.kt
```diff
@@ -15,6 +15,6 @@ class MapProjectionHelper {
     private val differenceLon = end.longitude - start.longitude
 
     fun toSmallMapLatLng(latLng: LatLng): LatLng {
```
Cliabhach: So...how is this different from just returning `latLng` directly?

PR #163, opened 2018-09-24
Greet returning users (AIC-465)

sam33rdhakal:
Updates:
* Moves membership functionality from info to membership module
* Greets the saved card holder in the welcome screen


welcome/src/main/kotlin/edu/artic/welcome/WelcomeFragment.kt
```diff
@@ -24,6 +24,7 @@ import edu.artic.viewmodel.Navigate
 import io.reactivex.Observable
 import io.reactivex.functions.Consumer
 import kotlinx.android.synthetic.main.fragment_welcome.*
+import timber.log.Timber
```
Cliabhach: We can remove the `Timber` import, no?

welcome/src/main/kotlin/edu/artic/welcome/WelcomeViewModel.kt
```diff
@@ -33,6 +34,7 @@ class WelcomeViewModel @Inject constructor(private val welcomePreferencesManager
                                            private val exhibitionDao: ArticExhibitionDao,
                                            generalInfoDao: GeneralInfoDao,
                                            languageSelector: LanguageSelector,
+                                           private val memberInfoPreferencesManager: MemberInfoPreferencesManager,
```
Cliabhach: Please move up two lines to be next to the other `private val`s.

welcome/src/main/kotlin/edu/artic/welcome/WelcomeViewModel.kt
```diff
@@ -115,6 +118,14 @@ class WelcomeViewModel @Inject constructor(private val welcomePreferencesManager
         shouldPeekTourSummary.onNext(welcomePreferencesManager.shouldPeekTourSummary)
     }
 
+    fun updateData() {
+        val activeCardHolder = memberInfoPreferencesManager.activeCardHolder
+        activeCardHolder?.let { cardHolder ->
```
Cliabhach: Add comment saying `this only works because we don't support logout.`

welcome/src/main/kotlin/edu/artic/welcome/WelcomeFragment.kt
sam33rdhakal: Yea removed.

welcome/src/main/kotlin/edu/artic/welcome/WelcomeViewModel.kt
sam33rdhakal: Done.

welcome/src/main/kotlin/edu/artic/welcome/WelcomeViewModel.kt
Cliabhach: This doesn't have to go in this file - anywhere appropriate, really.

sam33rdhakal: I don't think adding that comment here makes sense. The Sign-In feature in the app is only to validate the membership information that the user provides. Here we are just pulling the validated active card-holder info from the file.

Cliabhach: Hmm. I suppose? Either way that's more appropriate for the README.

PR #174, opened 2018-09-25
Audio tutorial (AIC-79)

sam33rdhakal:
Updates:
* Displays audio tutorial if the audio is being played for first time
* Fixes the status bar color issue
* Updates the `expires` string resource `[AIC-467]`


media_ui/src/main/java/edu/artic/media/ui/NarrowAudioPlayerViewModel.kt
```diff
@@ -1,10 +1,43 @@
...
+        audioServiceHook: AudioServiceHook,
+        private val audioPrefManager: AudioPrefManager
+) : NavViewViewModel<NarrowAudioPlayerViewModel.NavigationEndpoint>() {
+
+    /**
+     * Used to resume the audio player after user reads the audio tutorial.
+     */
+    val resumeAudioPlayBack: Subject<Boolean> = PublishSubject.create()
+
+    fun userSawAudioTutorial() {
```
Cliabhach: Move this function below `init`.

Cliabhach:
The Activity class does null checks on `window` - that might mean we might have a null value in `BaseFragment` under unusual scenarios.

PR #182, opened 2018-09-26
Hide Bottom Nav for Search & Details Screens (AIC-468)

caguilar187:


search/src/main/java/edu/artic/search/SearchAudioDetailFragment.kt
```diff
@@ -133,6 +133,9 @@ class SearchAudioDetailFragment : BaseViewModelFragment<SearchAudioDetailViewMod
                                 flags = Intent.FLAG_ACTIVITY_REORDER_TO_FRONT or Intent.FLAG_ACTIVITY_NO_ANIMATION
                             }
                             startActivity(mapIntent)
+                            if (requireActivity().javaClass.simpleName.equals("SearchActivity")) {
```
Cliabhach: Please add a function to `:navigation` module to do this check. `isInSearchActivity()` or something like that.

PR #187, opened 2018-09-27
Map Tutorial (AIC-78)

raigex:
Add header that shows up the first time a session that map is opened (except in certain conditions)
Add Tutorial fragment and flows


Cliabhach:
I think we'll need to get those translations in there soon. That ought to be its own PR, though.

PR #190, opened 2018-09-28
Fix appearance of ellipses in search results (AIC-461)

Cliabhach:
This switches field usage in the search results screen from `artist_display` (which has multiple lines) to `artist_title` (which is only one line).


details/src/main/kotlin/edu/artic/exhibitions/ExhibitionDetailViewModel.kt
```diff
@@ -69,8 +69,8 @@ constructor(dataObjectDao: ArticDataObjectDao,
                 .disposedBy(disposeBag)
 
         exhibitionObservable
-                .filter { it.legacy_image_mobile_url != null }
-                .map { it.legacy_image_mobile_url!! }
+                .filter { it.legacyImageUrl != null }
+                .map { it.legacyImageUrl!! }
```
sam33rdhakal: You could use `filterFlatMap` here.
`.filterFlatMap({ it.legacyImageUrl != null }, {it.legacyImageUrl!! })`

PR #194, opened 2018-09-28
Remove `zxing-android-embedded` dependency (AIC-459)

sam33rdhakal:
Updates:
- Copies the required `BarcodeEncoder` and removes `zxing-android-embedded` dependency

### DIFF
--- a/access_card/src/main/java/edu/artic/accesscard/barcode/BarcodeEncoder.kt
+++ b/access_card/src/main/java/edu/artic/accesscard/barcode/BarcodeEncoder.kt
@@ -1,0 +1,78 @@
+package edu.artic.accesscard.barcode
+
+
+import android.graphics.Bitmap
+
+import com.google.zxing.BarcodeFormat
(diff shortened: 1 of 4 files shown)


access_card/src/main/java/edu/artic/accesscard/barcode/BarcodeEncoder.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.google.zxing.common.BitMatrix
+
+
+/**
+ * Helper class for encoding barcodes as a Bitmap.
+ *
+ * Adapted from QRCodeEncoder, from the zxing project:
+ * https://github.com/zxing/zxing
+ *
+ * Licensed under the Apache License, Version 2.0.
```
Cliabhach: I think we might need to add a more explicit license statement here. Not sure. I'll bring it up with legal.

Cliabhach:
This also drops the camera permission, correct?

PR #197, opened 2018-09-28
Add the tiler.sh to version control

caguilar187:
Also load the default tile and use it for instances where it is required instead of loading from assets everytime

### DIFF
--- a/map/tiler.sh
+++ b/map/tiler.sh
@@ -1,0 +1,53 @@
+#! /bin/bash
+
+generateZoomLevels(){
+  FILENAME="$1"
+  ZOOM_LEVEL="$2"
+  FLOOR="$3"
+
+  SIZE=$((512*(2**$ZOOM_LEVEL)))
+
+  # should the scale value (currently 0.905) be played with modifcations need to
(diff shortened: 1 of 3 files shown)


map/tiler.sh
```diff
@@ -0,0 +1,53 @@
...
+
+generateZoomLevels(){
+  FILENAME="$1"
+  ZOOM_LEVEL="$2"
+  FLOOR="$3"
+
+  SIZE=$((512*(2**$ZOOM_LEVEL)))
+
+  # should the scale value (currently 0.905) be played with modifcations need to
+  # also occur in $LEFT_CHOP_AMOUNT_SCALED
```
Cliabhach: What does this comment mean?

map/tiler.sh
```diff
@@ -0,0 +1,53 @@
+#! /bin/bash
+
```
Cliabhach: Please add description to `README.md` explaining what this script is and (briefly) how to use it.

Cliabhach: Thanks!

PR #199, opened 2018-10-01
Proceed To Home Screen when Network Error (when cache is available)

sam33rdhakal:
Wrong execption type `NetworkErrorException` was used instead of `NetworkException`.

### DIFF
--- a/splash/src/main/kotlin/edu/artic/splash/SplashActivity.kt
+++ b/splash/src/main/kotlin/edu/artic/splash/SplashActivity.kt
@@ -1,6 +1,5 @@
 package edu.artic.splash
 
-import android.accounts.NetworkErrorException
 import android.app.AlertDialog
 import android.app.DialogFragment.STYLE_NO_FRAME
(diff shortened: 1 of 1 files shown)


splash/src/main/kotlin/edu/artic/splash/SplashActivity.kt
```diff
@@ -82,7 +82,7 @@ class SplashActivity : BaseViewModelActivity<SplashViewModel>(), TextureView.Sur
                             .setMessage(errorMessage)
                             .setPositiveButton(getString(android.R.string.ok)) { dialog, _ ->
                                 dialog.dismiss()
-                                if (it is NetworkErrorException) {
+                                if (it is NetworkException) {
```
Cliabhach: I do think we'll need a more extensive error-support in the near future. For now this will do.

Cliabhach:
The Circle CI error looks spurious. You can merge it in regardless.

PR #200, opened 2018-10-01
[Feature] Related tours (AIC-492)

sam33rdhakal:
Updates:
* Displays related tours for the currently playing `ArticObject`
* Starts tour when one of related tours is clicked.


media_ui/src/main/java/edu/artic/media/ui/AudioDetailsFragment.kt
```diff
@@ -143,6 +152,60 @@ class AudioDetailsFragment : BaseViewModelFragment<AudioDetailsViewModel>() {
...
+                    }
+                }
+                .disposedBy(disposeBag)
+    }
+
+    @UiThread
+    fun addRelatedToursToView(tours: List<ArticTour>) {
+        val textView = TextView(requireContext())
+        textView.text = resources.getString(R.string.relatedTours)
+        TextViewCompat.setTextAppearance(textView, R.style.SectionTitleWhite)
```
Cliabhach: I think this (the title/heading) should be defined in the XML. Would that be ok?

map/build.gradle
```diff
@@ -31,4 +31,5 @@ dependencies {
     implementation libs.arch_navigation_ui
     implementation libs.glide
     implementation libs.view_indicator
+    implementation project(':tour_manager')
```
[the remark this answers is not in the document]
sam33rdhakal: Moved. Also sorted the dependencies alphabetically.

media_ui/src/main/java/edu/artic/media/ui/AudioDetailsFragment.kt
sam33rdhakal: Moved it to layout file.

PR #201, opened 2018-10-01
Do not display location prompt if location permission is already granted (AIC-518)

sam33rdhakal:
Checks for permission before displaying the location permission prompt.

### DIFF
--- a/map/src/main/kotlin/edu/artic/map/MapFragment.kt
+++ b/map/src/main/kotlin/edu/artic/map/MapFragment.kt
@@ -18,10 +18,7 @@ import com.jakewharton.rxbinding2.view.clicks
 import com.jakewharton.rxbinding2.view.globalLayouts
 import com.jakewharton.rxbinding2.view.visibility
 import edu.artic.analytics.ScreenCategoryName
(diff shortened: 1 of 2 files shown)


map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -477,12 +471,17 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
...
-                            requireActivity().supportFragmentManager
-                                    ?.beginTransaction()
-
-                                    ?.replace(R.id.overlayContainer, LocationPromptFragment(), "LocationPromptFragment")
-                                    ?.addToBackStack("LocationPromptFragment")
-                                    ?.commit()
+                            /**
+                             * Display location prompt iff location permission is not granted.
+                             */
+                            if (requireActivity().hasFineLocationPermission()) {
```
Cliabhach: Shouldn't this be inverted?

sam33rdhakal: Nice catch. Updated it.

PR #206, opened 2018-10-02
Fix up search result visuals (AIC-462, AIC-464)

caguilar187:


search/src/main/java/edu/artic/search/SearchBaseFragment.kt
```diff
@@ -124,6 +128,14 @@ abstract class SearchBaseFragment<TViewModel : SearchBaseViewModel> : BaseViewMo
                             startActivity(mapIntent)
                             requireActivity().finish()
                         }
+                        SearchBaseViewModel.NavigationEndpoint.HideKeyboard -> {
+                            val imm = requireContext().getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+                            var view = requireActivity().currentFocus
```
Cliabhach: You should use our extension method. `requireActivity().hideSoftKeyboard()`.

PR #210, opened 2018-10-03
Fix the compass not animating by using SensorManager to pull in bearing (AIC-522)

caguilar187:


location/src/main/java/edu/artic/location/FuzedLocationProvider.kt
```diff
@@ -6,35 +6,83 @@ import android.app.Activity
...
+            Surface.ROTATION_270 ->
+                SensorManager.remapCoordinateSystem(defaultRotation, SensorManager.AXIS_MINUS_Y,
+                        SensorManager.AXIS_Z, correctedRotation)
+        }
+    }
+
     private val locationCallback: LocationCallback = object : LocationCallback() {
         override fun onLocationResult(locationResult: LocationResult?) {
             super.onLocationResult(locationResult)
             locationResult?.let {
```
Cliabhach: Replace with `locationResult?.lastLocation?.let {` since we don't really need anything else.

location/src/main/java/edu/artic/location/FuzedLocationProvider.kt
```diff
@@ -113,9 +169,21 @@ class FuzedLocationProvider(private val context: Context) : LocationProvider {
...
 
+    override fun onSensorChanged(event: SensorEvent) {
+        if (event.sensor.type == Sensor.TYPE_ACCELEROMETER) {
+            mGravity = event.values
+        } else if (event.sensor.type == Sensor.TYPE_MAGNETIC_FIELD) {
+            mGeomagnetic = event.values
+        }
+    }
+
+    override fun onAccuracyChanged(p0: Sensor?, p1: Int) {
```
Cliabhach: Please use same parameter names as defined in interface, replace body of method with ` = Unit`

PR #215, opened 2018-10-04
Audio details: use app language for dropdowns (AIC-382)

Cliabhach:
This covers the change in appearance of the dropdown/spinner view on the audio details screen. Previously, it showed the languages as spelled in that language (`English`, `Español`, `中文`), while now it'll show them all in the language of the application (`English`, `Spanish`, `Chinese`).

### DIFF
--- a/localization_ui/src/main/res/layout/fragment_language_settings.xml
+++ b/localization_ui/src/main/res/layout/fragment_language_settings.xml
@@ -114,7 +114,7 @@
                     android:layout_marginEnd="@dimen/marginDouble"
                     android:background="?attr/languageSettingsButtonBackground"
                     android:button="@null"
-                    android:text="@string/english"
+                    android:text="@string/english_in_that_language"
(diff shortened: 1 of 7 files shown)


localization_ui/src/main/res/layout/fragment_language_settings.xml
```diff
@@ -114,7 +114,7 @@
                     android:layout_marginEnd="@dimen/marginDouble"
                     android:background="?attr/languageSettingsButtonBackground"
                     android:button="@null"
-                    android:text="@string/english"
+                    android:text="@string/english_in_that_language"
```
caguilar187: This isn't the dropdown view, should we be changing this view?

Cliabhach: This requires a different resource. Since I redefined `@string/english`, `@string/spanish`, and `@string/chinese`, this was the term I used for the untranslatable text. I can rename to `@string/english_in_english` etc. if that helps?

PR #222, opened 2018-10-08
Add missing accessibility labels for Search, Play/Pause, Map Annotations, Close, etc. (AIC-526)

caguilar187:


db/src/main/kotlin/edu/artic/db/Accessiblable.kt
```diff
@@ -0,0 +1,8 @@
+package edu.artic.db
+
+/**
+ * Used to ensure content that requires content description can use this interface
+ */
+interface Accessiblable {
```
Cliabhach: Rename to `HasContentDescription`, maybe? `AccessibilityAware`? I'm not keen on `Accessiblable`.

db/src/main/kotlin/edu/artic/db/Accessiblable.kt
```diff
@@ -0,0 +1,8 @@
+package edu.artic.db
+
+/**
+ * Used to ensure content that requires content description can use this interface
```
Cliabhach: Change comment to

    Used to provide a reasonable content description to accessibility services

PR #225, opened 2018-10-09
Navigation issues, between Search and the main activities, and between home and secondary tabs

caguilar187:
[AIC-539] Hide Audio Guide Translation If Only 1 Available
[AIC-525] Tapping Category Tap Should Clear Screen History
[AIC-498] Selecting Search Icon Should Display Keyboard
[AIC-528] Increase default map tilt angle, and straighten
Ensure a locale is set on first launch


caguilar187:
@Cliabhach it could but it serves little point since there is no second level nav on either of them. At the very least that is true on audio. Map maybe it should give the prompt to exit a tour?

PR #227, opened 2018-10-09
Address some home-screen concerns relevant to smaller screens (AIC-401)

Cliabhach:
This also moves `DateTimeHelper.kt` to the localization module, in keeping with its language-dependent behaviors.


Cliabhach:
I'll rebase on `dev` first.

PR #230, opened 2018-10-09
All Tours screen style fixes (AIC-403)

Cliabhach:
This includes improvements primarily to the 'all tours' layouts, but it also has a couple nice benefits for toolbars throughout the application.

The application default font is now `ideal_sans_book`.


base/src/main/res/values/styles.xml
```diff
@@ -7,6 +7,7 @@
         <item name="colorPrimary">@color/colorPrimary</item>
         <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
         <item name="colorAccent">@color/colorAccent</item>
+        <item name="android:fontFamily">@font/ideal_sans_medium</item>
```
sam33rdhakal: I think we use medium fonts for title, subtitle, buttons etc where the text should be bold. IMP we should use regular font here, maybe ideal_sans_book.

Cliabhach: I've made the necessary change.

PR #231, opened 2018-10-09
Include link to website when no search results are found (AIC-515)

sam33rdhakal:
* Adds correct spacing between empty result prompt and title (AIC-515)
* underlines `visit our website` string
* Loads website when empty search result prompt is clicked. (AIC-534)


Cliabhach:
I think we could push much of that logic out of `SearchBaseViewModel` and into the subclasses. That's an issue for another ticket, though.

PR #243, opened 2018-10-11
Dont pop inclusive otherwise the navcontroller loses its state [AIC-547] 

caguilar187:


Cliabhach:
Does this fix the observed crash?

PR #246, opened 2018-10-11
Increase white margin around selected marker's pin (AIC-508)

caguilar187:


map/src/main/kotlin/edu/artic/map/rendering/ObjectsMapItemRenderer.kt
```diff
@@ -72,19 +85,35 @@ class ObjectsMapItemRenderer(private val objectsDao: ArticObjectDao)
...
+            existing: MarkerHolder<MapItemModel>,
+            selectedArticObject: ArticObject?,
+            visibleRegion: VisibleRegion,
+            mapChangeEvent: MapChangeEvent) {
         val position = getAdjustedLocationFromItem(existing.item)
         val meta = existing.marker.metaData()
-                ?: MarkerMetaData(existing.item, loadedBitmap = false, requestDisposable = null)
+                ?: MarkerMetaData(existing.item, loadedBitmap = false, requestDisposable = null, isSelected = false)
+
+        val isSelected = existing.item.isSelected(selectedArticObject)
```
Cliabhach: If we rename local `val` to `newlySelected` does that help readability?

map/src/main/kotlin/edu/artic/map/rendering/ObjectsMapItemRenderer.kt
```diff
@@ -72,19 +85,35 @@ class ObjectsMapItemRenderer(private val objectsDao: ArticObjectDao)
...
+                || (position.isCloseEnoughToCenter(visibleRegion.latLngBounds) || isSelected)) {
             if (!meta.loadedBitmap) {
                 // show loading while its loading.
                 existing.marker.setIcon(loadingBitmap)
 
                 // enqueue replacement and add to marker meta.
                 existing.marker.tag = meta.copy(
-                        requestDisposable = enqueueBitmapFetch(item = existing.item, mapChangeEvent = mapChangeEvent))
+                        requestDisposable = enqueueBitmapFetch(item = existing.item, mapChangeEvent = mapChangeEvent), isSelected = isSelected)
+            } else if (!meta.isSelected && isSelected) {
```
Cliabhach: Isn't this `if` block doing the same as the next one? You could use the `kotlin.Boolean` infix `xor` operator.

map/src/main/kotlin/edu/artic/map/rendering/ObjectsMapItemRenderer.kt
```diff
@@ -72,19 +85,35 @@ class ObjectsMapItemRenderer(private val objectsDao: ArticObjectDao)
...
         val position = getAdjustedLocationFromItem(existing.item)
         val meta = existing.marker.metaData()
-                ?: MarkerMetaData(existing.item, loadedBitmap = false, requestDisposable = null)
+                ?: MarkerMetaData(existing.item, loadedBitmap = false, requestDisposable = null, isSelected = false)
+
+        val isSelected = existing.item.isSelected(selectedArticObject)
 
         if (mapChangeEvent.displayMode !is MapDisplayMode.CurrentFloor
-                || position.isCloseEnoughToCenter(visibleRegion.latLngBounds)) {
+                || (position.isCloseEnoughToCenter(visibleRegion.latLngBounds) || isSelected)) {
```
Cliabhach: Can we drop the parentheses and move `|| isSelected) {` onto its own line?

PR #262, opened 2018-10-15
Updates the size of playback control icons 

sam33rdhakal:
As the title describes, this pull req is for 
* Updating the playback icon in search details screen
* Updates the color of the close icon.
* Updates the margin between buttons in access member card screen


sam33rdhakal:
> Please rename the close drawable, as it is no longer black. I personally like `ic_close_white` but you can choose something else.

Renamed it to `ic_close_white` as you suggested.

PR #263, opened 2018-10-15
Audio lookup appearance on large devices (AIC-416)

Cliabhach:
Another PR relating to AIC-416 here. This one increases the horizontal subtitle margin to 50dp on all but the smallest devices and centers the number pad vertically in its layout.


audio/src/main/res/layout/fragment_audio_lookup.xml
```diff
@@ -110,7 +110,7 @@
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@id/subheader"
         app:layout_constraintBottom_toTopOf="@id/newPlayer"
-        app:layout_constraintVertical_bias="0"
+        app:layout_constraintVertical_bias="0.5"
```
sam33rdhakal: I think the default value of vertical_bias is `0.5` so this is not necessary here. Also according to ticket AIC-416, the visibility of audioPlayer should not change the position of the keypad. So, we should add `guideline` or `space` equal to the height of `narrow audio player` and bottom constraint keypad to that view.

Cliabhach: Understood. I'll get on that.

PR #267, opened 2018-10-15
Update to SDK 28, fix dependencies and use lineheight

caguilar187:
This covers the following changes:
* Upgrade `target` and `compile` sdks from `27` (Oreo) to `28` (Pie).
* Upgrade Exoplayer media dependency
*# from `2.8.2` to `2.9.0`
*# suppress attempts thereof to bring in older support lib dependencies
* Add explicit network security configuration
*# This currently forbids cleartext (`http://` instead of `https://`) traffic to all hosts by default
*# One exception is in place for `artic.edu`, to be removed once we receive confirmation from the server team
* Use the new `app:lineHeight` attribute to enforce specific text stylings on certain `event`- and `exhibition`-related layouts

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ buildscript {
         moshi_version = '1.6.0'
         room_version = '1.1.1'
         glide_version = '4.4.0'
-        firebase_version = '16.0.1'
+        firebase_version = '16.0.4'
         crashlytics_version = '2.9.5'
         exo_player_version = '2.9.0'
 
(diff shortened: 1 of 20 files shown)


build.gradle
```diff
@@ -86,6 +87,15 @@ buildscript {
 }
 
 subprojects {
+
+    configurations.all {
+        resolutionStrategy {
+            force   "com.android.support:support-v4:${support_version}",
+                    "com.android.support:support-media-compat:${support_version}",
```
Cliabhach: Please use our pre-existing `media_compat` and `v4_support` constants.

PR #276, opened 2018-10-16
Updates correct event category for analytics  - (AIC-552)

sam33rdhakal:
`ScreenCategoryName` was used as event category instead of using `EventCategoryName` for analytics events. Changes in this pull req will fix those issues.


sam33rdhakal:
@Cliabhach `ScreenCategoryName` is used for tracking the screen view whereas `EventCategoryName` is used for categorizing the analytics events. `ScreenCategoryName` is a bad name, we don't track screen by categories, we could rename it to `ScreenName`.

Cliabhach:
Interesting. Would you be all right renaming it accordingly as part of this PR?

sam33rdhakal:
@Cliabhach Renamed the Class.

PR #277, opened 2018-10-16
Ensure the map's session remains alive for the duration of the app being in use

caguilar187:


location/src/main/java/edu/artic/location/LocationUtils.kt
```diff
@@ -24,6 +24,10 @@ val museumBounds: LatLngBounds = LatLngBounds(
  */
 val centerOfMuseumOnMap = LatLng(41.879592, -87.622491)
 
+const val distanceFromCenter = 250.0f
```
Cliabhach: Rename to `museumBoundsRadiusInMeters`.

location/src/main/java/edu/artic/location/LocationUtils.kt
```diff
@@ -24,6 +24,10 @@ val museumBounds: LatLngBounds = LatLngBounds(
...
+
 fun isLocationInMuseum(location : Location) : Boolean {
-    return museumBounds.contains(LatLng(location.latitude, location.longitude))
+    val distance = FloatArray(1)
+    Location.distanceBetween(location.latitude, location.longitude, centerOfMuseumOnMap.latitude, centerOfMuseumOnMap.longitude, distance)
```
Cliabhach: Shouldn't we be using `museumBounds.center` instead of `centerOfMuseumOnMap`?

map/src/main/kotlin/edu/artic/map/MapActivity.kt
```diff
@@ -85,4 +88,15 @@ class MapActivity : BaseActivity() {
         Glide.get(this).setMemoryCategory(MemoryCategory.NORMAL)
     }
 
+    override fun onBackPressed() {
+        if (supportFragmentManager.backStackEntryCount == 0) {
+            if (navController.currentDestination?.id == R.id.mapFragment) {
+                val intent = NavigationConstants.HOME.asDeepLinkIntent()
```
Cliabhach: Please simplify the three lines here and in the other `onBackPressed` overrides using a function in the `navigation` module. Something like

    fun linkHome(): Intent

PR #278, opened 2018-10-16
[AIC-570] add support for keystore to configs

caguilar187:


app/build.gradle
```diff
@@ -30,6 +30,16 @@ android {
         buildConfigField "String", "VERSION_NAME_FOR_DISPLAY", "VERSION_NAME"
         resValue "string", "com.crashlytics.android.build_id", "$majorAndMinorVersion.$patchVersionCode"
     }
+
+    signingConfigs {
+        release {
+            storeFile file(envVariable("keystore",true))
```
Cliabhach: Can you check whether

    storeFile file(envVariable("keystore",true) ?: "DUMMY")

would work?

PR #279, opened 2018-10-16
Prevent crashes caused by misconfiguration

Cliabhach:
If the '.env' files are not set up right, or for that matter not defined at all, the app in its current state will crash at startup.

With these commits, a more manageable error message will be shown instead. It also adds a nice affordance to bundle a sample JSON resource with the apk, if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,24 +46,28 @@ The project maintains an overview of the more interesting modules in
 
 ## How to Build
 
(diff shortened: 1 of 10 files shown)


Cliabhach:
@caguilar187 As you requested offline, I will remove the gradle changes from this PR.

Cliabhach:
Thanks! I'll try to get the Circle CI build to succede.

PR #281, opened 2018-10-17
Use translations for known values in Event Details

caguilar187:


details/src/main/kotlin/edu/artic/events/EventDetailFragment.kt
```diff
@@ -124,6 +125,19 @@ class EventDetailFragment : BaseViewModelFragment<EventDetailViewModel>() {
 
 
         viewModel.eventButtonText
+                .filter { it==("Buy Tickets")}
```
Cliabhach: Isn't this covered by the other `eventButtonText` stream, just below?

details/src/main/kotlin/edu/artic/events/EventDetailFragment.kt
```diff
@@ -25,6 +25,7 @@ import edu.artic.viewmodel.BaseViewModelFragment
 import edu.artic.viewmodel.Navigate
 import io.reactivex.rxkotlin.Observables
 import kotlinx.android.synthetic.main.fragment_event_details.*
+import kotlinx.android.synthetic.main.fragment_exhibition_details.view.*
```
Cliabhach: I don't think we need this import. Will it work without?

details/src/main/kotlin/edu/artic/events/EventDetailFragment.kt
```diff
@@ -20,7 +20,6 @@ import edu.artic.db.models.ArticEvent
 import edu.artic.details.R
 import edu.artic.image.GlideApp
 import edu.artic.image.ImageViewScaleInfo
-import edu.artic.image.listenerAnimateSharedTransaction
```
Cliabhach: Aren't we still using that on line 91?

PR #288, opened 2018-10-17
[AIC-575] change description text to floor of exhibition

caguilar187:


map/src/main/kotlin/edu/artic/map/SearchObjectDetailsViewModel.kt
```diff
@@ -16,6 +16,7 @@ import io.reactivex.rxkotlin.withLatestFrom
 import io.reactivex.subjects.BehaviorSubject
 import io.reactivex.subjects.PublishSubject
 import io.reactivex.subjects.Subject
+import kotlinx.android.synthetic.main.fragment_map.view.*
```
Cliabhach: Do we need this?

PR #291, opened 2018-10-18
WIP: Welcome: Standardize usage of 'Cell' ViewModels

Cliabhach:
This mostly just removes some duplicate code. There's also some better documentation of the various `All` fragments.

This PR is on hold until #297 and #299 are merged in. It will need some relatively-trivial updates after that point to account for moved and deleted files.


tours/src/main/kotlin/edu/artic/tours/AllToursViewModel.kt
```diff
@@ -67,9 +67,9 @@ class AllToursViewModel @Inject constructor(
 }
 
 /**
- * This class is fundamentally the same as `WelcomeTourCellViewModel` in the :welcome module.
+ * ViewModel responsible for building each item in the tour summary list.
  */
-class AllToursCellViewModel(
+class TourCellViewModel(
```
caguilar187: let's move these into their own files now that they are being used in both places

exhibitions/src/main/kotlin/edu/artic/exhibitions/AllExhibitionsFragment.kt
```diff
@@ -50,6 +47,16 @@ class AllExhibitionsFragment : BaseViewModelFragment<AllExhibitionsViewModel>()
 
     override fun setupBindings(viewModel: AllExhibitionsViewModel) {
         val adapter = (recyclerView.adapter as AllExhibitionsAdapter)
+
+        /* Ensure search events go through ok. */
+        searchIcon
```
caguilar187: can you seperate this out into it's own pull request, i believe AllTours and AllEvents do the same thing so we should update those as well

exhibitions/src/main/res/layout/fragment_all_exhibitions.xml
```diff
@@ -7,32 +7,38 @@
     android:layout_height="match_parent"
     android:animateLayoutChanges="true">
 
-    <android.support.design.widget.AppBarLayout
-        android:id="@+id/appBarLayout"
+    <android.support.v7.widget.Toolbar
```
caguilar187: similar to above lets create a layout we include for this and use in the three fragments. At least for that PR. After that we can use it in other places as well

welcome/src/main/kotlin/edu/artic/welcome/WelcomeEventsAdapter.kt
```diff
@@ -14,8 +15,8 @@ import kotlinx.android.synthetic.main.welcome_event_cell_layout.view.*
  * @author Sameer Dhakal (Fuzz)
  */
 
-class WelcomeEventsAdapter : AutoHolderRecyclerViewAdapter<WelcomeEventCellViewModel>() {
-    override fun View.onBindView(item: WelcomeEventCellViewModel, position: Int) {
+class WelcomeEventsAdapter : AutoHolderRecyclerViewAdapter<EventCellViewModel>() {
```
caguilar187: since you are using the EventCellViewModel at this point and the cells are the same between both. Can we swap over to the EventsAdapter in the events package? The only difference is items and thats control in the WelcomeViewModel

caguilar187 (changes requested):

tours/src/main/kotlin/edu/artic/tours/AllToursViewModel.kt
Cliabhach: Good idea.

exhibitions/src/main/kotlin/edu/artic/exhibitions/AllExhibitionsFragment.kt
Cliabhach: Those other two already use this mechanism - AllExhibitions was the last hold-out. I'll create a PR for that and the layout pieces mentioned below.

Cliabhach: It's up! Check out #297.

welcome/src/main/kotlin/edu/artic/welcome/WelcomeEventsAdapter.kt
Cliabhach: That change will need to be covered as a separate part of #298, so I'll close this particular comment as resolved within the scope of #291.

exhibitions/src/main/res/layout/fragment_all_exhibitions.xml
Cliabhach: This is done too! It's in commit e68b240b, merged in #299.

PR #297, opened 2018-10-19
Standardize styling of the various search icons

Cliabhach:
This does not significantly modify the layout of any such `ImageView`; note that this work is primarily intended to lay the groundwork for tasks related to #298 (right now, PR #291 and PR #299 are the best examples of that).


Cliabhach:
@caguilar187 If you have a moment today, I'd appreciate a quick look at this. It shouldn't be terribly controversial.

PR #305, opened 2018-10-24
Map artwork ordering (AIC-586)

sam33rdhakal:
We were expecting database would preserve the order of ids for the following sql query.

`select * from ArticObject where nid in (:ids)`

Database does not preserve the order of ids given in`IN` clause.
This PR fixes it by sorting the database result based on the input `idList`

### DIFF
--- a/search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
+++ b/search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
@@ -8,6 +8,8 @@ import edu.artic.analytics.AnalyticsTracker
 import edu.artic.analytics.EventCategoryName
 import edu.artic.db.daos.ArticDataObjectDao
 import edu.artic.db.daos.ArticGalleryDao
+import edu.artic.db.daos.ArticObjectDao
+import edu.artic.db.daos.ArticSearchObjectDao
 import edu.artic.db.models.*
 import edu.artic.viewmodel.NavViewViewModel
 import edu.artic.viewmodel.Navigate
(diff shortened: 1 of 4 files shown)


Cliabhach:
What do you think of adding an order by nid clause to the query? That might help the performance of the `find` operation and make it clear that the results are not sorted.

search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
```diff
@@ -174,4 +176,36 @@ open class SearchBaseViewModel @Inject constructor(
...
+                     * Database does not preserve the order of ids.
+                     */
+                    objectDao.getObjectsByIdList(idsList)
+                            .toObservable()
+                            .map { unSortedObjects ->
+                                /**
+                                 * Sorting the objects based on idsList
+                                 */
+                                val sortedObjects = mutableListOf<ArticObject>()
+                                for (id in idsList) {
```
Cliabhach: I think this would look better with the Kotlin Stream operators, as

    val sortedObjects: List<ArticObject> = idsList.mapNotNull { id ->
        unSortedObjects.find { it.nid == id }
    }

or even

    return@map idsList.mapNotNull { id ->
        unSortedObjects.find { it.nid == id }
    }

Then you can make the function return an `Observable<List<ArticObject>>`. It'll match the type of `suggestedArtworks` too that way. Iirc `mapNotNull` preserves order the same way `for...in` does.

search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
```diff
@@ -174,4 +176,36 @@ open class SearchBaseViewModel @Inject constructor(
...
+     */
+    protected fun getSuggestedArtworks(searchSuggestionsDao: ArticSearchObjectDao,
+                                       objectDao: ArticObjectDao
+    ): Observable<MutableList<ArticObject>> {
+        return searchSuggestionsDao.getDataObject()
+                .toObservable()
+                .map { suggestedSearchOptions -> suggestedSearchOptions.searchObjects }
+                .flatMap { idsList ->
+                    /**
+                     * Database does not preserve the order of ids.
```
Cliabhach: Put this comment on the database method too.

search/src/main/java/edu/artic/search/SearchBaseViewModel.kt
```diff
@@ -174,4 +176,36 @@ open class SearchBaseViewModel @Inject constructor(
     open fun onClickSeeAll(header: Header) {
 
     }
+
+    /**
+     * Returns suggested artworks.
```
Cliabhach: Perhaps `Returns suggested artworks, in order.`? I'm kind ambivalent about which we go for here.

Cliabhach:
No trouble. Feel free to merge it in whenever.

PR #306, opened 2018-10-24
Preserve the order of tours on `All Tours` screen (AIC-587)

sam33rdhakal:
Instead of sorting tour by its title, this PR preserves the order of tours in CMS.


Cliabhach:
I've rebased on `dev`. If Circle CI functions properly this time, we'll merge it in.

PR #307, opened 2018-10-25
Membership backstack adjustments (AIC-580)

Cliabhach:
This PR ensures that the link from the 'home' section to the member card screen may always be reversed with a single operation. When testing, make sure to consider situations where the 'info' section is already showing the member card and thus does not (at first glance) need to navigate anywhere.


caguilar187:
Testing one scenarios prior to merging

Cliabhach:
Sure. We can pick this up for discussion further on Monday the 29th, if that works for you.

caguilar187:
Not changing current behaviour

PR #309, opened 2018-10-25
Correct the panning behavior of the map when selecting tour stops (AIC-585, AIC-579)

caguilar187:
This PR also, fix a bug where the intro tour stop isn't animated to just the selected stops.

Additionally we add an initial zoom that is slightly closer so that the departments are already loaded on the map and matching iOS

### DIFF
--- a/db/src/main/kotlin/edu/artic/db/models/ArticTour.kt
+++ b/db/src/main/kotlin/edu/artic/db/models/ArticTour.kt
@@ -164,5 +164,5 @@ fun ArticTour.getIntroStop(): ArticTour.TourStop {
 }
 
 fun ArticTour.TourStop.isIntroStop(): Boolean {
-    return this.objectId === INTRO_TOUR_STOP_OBJECT_ID
+    return this.objectId == INTRO_TOUR_STOP_OBJECT_ID
 }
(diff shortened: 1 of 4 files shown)


map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -343,18 +344,18 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
...
                     } else {
                         // Make sure we can see all of the specified positions
-                        map.moveCamera(CameraUpdateFactory
-                                .newLatLngBounds(
-                                        LatLngBounds.builder()
-                                                .includeAll(bounds)
-                                                .build(),
-                                        0
-                                ))
+                        val lanLngBounds = LatLngBounds.builder().includeAll(bounds).build()
```
Cliabhach: uh. lanLngBounds? What do you think of making this into a `val` for the center? Or rename to `areaOfInterestBounds`?

db/src/main/kotlin/edu/artic/db/models/ArticTour.kt
```diff
@@ -164,5 +164,5 @@ fun ArticTour.getIntroStop(): ArticTour.TourStop {
 }
 
 fun ArticTour.TourStop.isIntroStop(): Boolean {
-    return this.objectId === INTRO_TOUR_STOP_OBJECT_ID
+    return this.objectId == INTRO_TOUR_STOP_OBJECT_ID
```
Cliabhach: **_nice_**.

map/src/main/kotlin/edu/artic/map/MapFragment.kt
Cliabhach: That works!

PR #310, opened 2018-10-26
Use the official `okhttp3-downloadprogress-interceptor` library

Cliabhach:
The official library may be seen at https://github.com/jobinlawrance/okhttp3-downloadprogress-interceptor.

By merging this, we fulfill another part of our licensing obligations without effecting changes on runtime functionality.


Cliabhach:
The `:db` module has some otherwise-superfluous gradle lines to enable `@Parcelize` support. I wonder if those might be interfering with the build?

Cliabhach:
Nope, it was just a minor `implementation`/`api` configuration distinction causing trouble. Maybe I should file a bug regarding the error message?

PR #285, opened 2018-10-17
Safeguard unit test threading

Cliabhach:
Ideally, this will remove some flakiness from our tests in `AppDataManagerTest.kt`.

Said uncertainty was introduced by 9d41f309997f65.

This PR also reduces the memory usage of the JVM build process (advice derived from https://discuss.circleci.com/t/circleci-2-0-android-with-kotlin/17531/2 ) and switches the version of the Android Gradle Plugin in use to the latest stable release (v3.2.1).

PR #311, opened 2018-10-27
Update gradle dependencies for October 2018

Cliabhach:
This is the first in what will hopefully be a series of regular maintenance changesets. This deals primarily with standardizing the 27 `build.gradle` files currently included in the repository.

While the functionality of output APKs should not be different, with any luck these changes will shed some light onto our ongoing build instability. Specific fixes for that issue are being developed at #285.

Included in this PR:
1. Update Kotlin Libs from JDK 7 variant to JDK 8 variant
2. Remove duplicate `apply plugin: X` lines from 26 `build.gradle` files
    - Here's a breakdown of the duplicates:
        - `com.android.library` (x1)
        - `kotlin-android` (x2)
        - `kotlin-android-extensions` (x24)
        - `kotlin-kapt` (x25)
    - Of these, only `kotlin-android-extensions` was not applied by the root `build.gradle`, so that had to be done as well
3. Use the same sorting logic for all of the `dependencies` blocks, with three primary goals
    1. Make it easier to see what each module is actually bringing in
    2. Reduce the chance of merge conflicts if two different contributors need to modify the same gradle configs
    3. Call out frequently-used modules for inclusion in [MODULES.md](MODULES.md)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,10 +106,13 @@ subprojects {
         apply plugin: 'com.android.library'
     }
     apply plugin: 'kotlin-android'
-    if (name != 'analytics' && name != 'image') {
-        // Most modules make use of the synthetic accessors; ':db' uses the experimental 'Parcelize' annotation, so it does some configuration beyond that.
-        apply plugin: 'kotlin-android-extensions'
-    }
+    // Most modules make use of the synthetic 'findViewById' accessors, part of
+    // the 'kotlin-android-extensions' plugin. The ':db' module goes beyond, as
+    // it also takes advantage of the experimental 'Parcelize' annotation.
+    //
+    // Any module that needs experimental features should enable such in the
+    // associated build.gradle, not here.
+    apply plugin: 'kotlin-android-extensions'
     apply plugin: 'kotlin-kapt'
 
     android {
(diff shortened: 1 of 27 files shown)


build.gradle
```diff
@@ -106,6 +106,10 @@ subprojects {
         apply plugin: 'com.android.library'
     }
     apply plugin: 'kotlin-android'
+    if (name != 'analytics' && name != 'image') {
+        // Most modules make use of the synthetic accessors; ':db' uses the experimental 'Parcelize' annotation, so it does some configuration beyond that.
+        apply plugin: 'kotlin-android-extensions'
```
caguilar187: i know these two don't need it but is there harm in applying them? looking at this from a simplification stand point

Cliabhach: Slightly, yes. Since I'm
1. trying to avoid changing functionality here
2. still not confident I know precisely what the plugin does

I scheduled that for a subsequent PR. For a point of reference, the first time you apply the kapt plugin to a module you add at least .5 seconds to the configuration time.

Cliabhach:
I'll rebase on dev and add a few commits I forgot to push up.

PR #313, opened 2018-10-29
Map artwork ordering  (AIC-584)

sam33rdhakal:
Updates:
* Brings the selected tour stop marker to the front when overlapped
* Focuses the intro marker object when tour carousel is displaying tour intro object


map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -196,6 +196,24 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
              * locks us into just that area
              */
             setLatLngBoundsForCameraTarget(mapDisplayBounds)
+
+            /**
+             * Marker.showInfoWindow() method is used (@see MapViewModel.retrieveMarkerByObjectId)
+             * to bring the selected marker to front (markers are sometime overlapped by other
```
Cliabhach: I think the inline link style `(see [MapViewModel.retrieveMarkerByObjectId])` would be better here. Also, how about `markers sometimes overlap` for the parenthetical on the next line?

sam33rdhakal: Agreed. Updated the documentation.

PR #315, opened 2018-10-30
Hides subtitle for tour search cells (AIC-596)

sam33rdhakal:


search/src/main/java/edu/artic/search/SearchResultsAdapter.kt
```diff
@@ -76,6 +77,7 @@ class SearchResultsAdapter : AutoHolderRecyclerViewAdapter<SearchBaseCellViewMod
 
                     }
                     is SearchTourCellViewModel -> {
```
Cliabhach: Please remove the empty `is` branches and add an `else` block setting `itemSubTitle` to `VISIBLE`.

sam33rdhakal: Updated.

PR #316, opened 2018-10-30
Account for weightless tours (AIC-597)

Cliabhach:
Some ArticTour objects may not come with a valid `weight` property. Until we derive a more generic solution for unexpected data formation, this will prevent issues at application startup caused by not detecting this property.


Cliabhach:
Oh, I see. I broke one of the tests with that update to `::enforceSanityCheck`. Now that the commit has been edited to mock out that SQLiteOpenHelper, it should be able to pass on Circle CI.

Cliabhach:
And it did! @sam33rdhakal would you re-approve this?

Cliabhach:
We'll need to make use of `weight` too, now that we have it coming in correctly.

PR #317, opened 2018-10-30
Advances tour stop when it's audio playback ends (AIC-555)

sam33rdhakal:
* Advances tour stop when selected stop's audio translation playback ends.

### DIFF
--- a/tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
+++ b/tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
@@ -1,6 +1,7 @@
 package edu.artic.tours.manager
 
 import android.annotation.SuppressLint
+import android.support.annotation.WorkerThread
 import com.fuzz.rx.Optional
 import com.fuzz.rx.filterValue
 import edu.artic.db.INTRO_TOUR_STOP_OBJECT_ID
@@ -8,9 +9,11 @@ import edu.artic.db.daos.ArticAudioFileDao
 import edu.artic.db.models.ArticTour
 import edu.artic.db.models.AudioFileModel
 import io.reactivex.rxkotlin.Observables
+import io.reactivex.rxkotlin.subscribeBy
 import io.reactivex.subjects.BehaviorSubject
 import io.reactivex.subjects.PublishSubject
 import io.reactivex.subjects.Subject
+import timber.log.Timber
 
 /**
  * Responsible for managing the communication between tour carousel and the map.
@@ -32,6 +35,7 @@ class TourProgressManager(val audioFileDao: ArticAudioFileDao) {
      * If [audioFileModel] belongs to currently selected tour i.e. [selectedTour],
      * this method advances [selectedStop] to next tour stop.
      */
+    @WorkerThread
     @SuppressLint("CheckResult")
     fun playBackEnded(audioFileModel: AudioFileModel) {
         Observables.combineLatest(selectedTour.filterValue(), selectedStop)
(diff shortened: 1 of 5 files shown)


tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
```diff
@@ -23,4 +27,57 @@ class TourProgressManager {
     val selectedTour: Subject<Optional<ArticTour>> = BehaviorSubject.createDefault(Optional(null))
     val proposedTour: Subject<Optional<Pair<ArticTour, ArticTour.TourStop>>> = BehaviorSubject.createDefault(Optional(null))
     val leaveTourRequest: Subject<Boolean> = PublishSubject.create()
+
+    /**
+     * If [audioFileModel] belongs to currently selected tour i.e. [selectedTour],
+     * this method advances [selectedStop] to next tour stop.
+     */
+    @SuppressLint("CheckResult")
+    fun playBackEnded(audioFileModel: AudioFileModel) {
```
Cliabhach: Please add a `@WorkerThread` annotation to make it clear this shouldn't run on the main thread.

tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
```diff
@@ -23,4 +27,57 @@ class TourProgressManager {
...
+                     */
+                    val audioFileID: String? = if (currentStopID == INTRO_TOUR_STOP_OBJECT_ID) {
+                        tour.tourAudio
+                    } else {
+                        tour.tourStops.find { it.objectId == currentStopID }?.audioId
+                    }
+
+                    audioFileID?.let { audioID ->
+                        audioFileDao
+                                .getAudioByIdAsync(audioID)
```
Cliabhach: What do you think of making `getAudioByIdAsync` return `Observable` instead of `Flowable`? That way we won't need the `::toObservable` calls here and in `TourCarousalIntroViewModel`.

tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
```diff
@@ -23,4 +27,57 @@ class TourProgressManager {
...
+                                .toObservable()
+                                .subscribe { audioFile ->
```
Cliabhach: Please add a simple `onError` to this subscription, on the off chance that `audioID` is malformed. Something that does nothing in production but uses Timber in debug, perhaps?

tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
sam33rdhakal: Room database does not support `Observable`. Please see https://developer.android.com/training/data-storage/room/accessing-data#query-observable
 
Leaving this as it is.

tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
sam33rdhakal: Good idea. Updated.

tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
sam33rdhakal: Added.

tour_manager/src/main/java/edu/artic/tours/manager/TourProgressManager.kt
Cliabhach: Ah, that support was added to Room 2.0.0 back in july (see note in https://medium.com/androiddevelopers/room-rxjava-acb0cd4f375). We can take advantage of it whenever we upgrade to that or a higher version.

PR #321, opened 2018-11-01
Formalise choice of gradle dependencies

Cliabhach:
Some of the app's explicit dependencies were being superseded at build time - this PR establishes the _actual_ numbers in use. The official term is dependency resolution, or something like that.

As an example: say we requested RxJava 2.1.7 and RxKotlin 2.3.0. The RxJava dependency would be changed to 2.2.0, as that is the most recent version supported by RxKotlin 2.3.0.


Cliabhach:
Rebased on dev, should be easy to merge in now.

PR #322, opened 2018-11-01
Bump versions for 1.1.0 (AIC-601)

Cliabhach:
Not much here, really. Version name to 1.1.0, version code to 1001.


Cliabhach:
After some minor discussion, we determined that this is really more of a 1.0.1 release.

PR #323, opened 2018-11-02
Fixes object marker click issues (AIC-605)

sam33rdhakal:
`InfoWindowAdapter` was implemented incorrectly. Hidden InfoWindow was intercepting the clicks. 
Don't know why view initialized as `View(requireContext())` taking large space. When it is replaced with 
```
TextView(requireContext()).apply {
                        text = " "
                    }
```
it worked fine.


Cliabhach:
Perhaps we should set a maximum text length of e.g. 1 too?

PR #326, opened 2018-11-02
Prepping for release of 1.0.1 to store

Cliabhach:
Previous release: #301.

This contains all of the work made on `dev` since commit `82683edcdaafd69f`. The merge commit should be tagged accordingly. @caguilar187, can you review?


Cliabhach:
I've merged this from the command line. That build has already been released, so there's no point dwelling on formality.

PR #327, opened 2018-11-05
Map: Focuses the correct tour stop marker (AIC-584)

sam33rdhakal:
When the map is loaded in tour mode with specific tour stop, that tour stop needs to be focused. This PR adds the feature to tag the marker (the one to be focused) as the selected marker and then focus it later when created.
Fixes both AIC-584 and AIC-595.


sam33rdhakal:
Sure. We need to simplify the map renderers and add more documentation.

PR #328, opened 2018-11-06
Use separate ViewHolders in language dropdown adapter (AIC-590)

sam33rdhakal:
Now `LanguageAdapter` is responsible for creating both dropdown cells and spinner cell. 
No need to generate the background detecting the layout changes.


media_ui/src/main/res/drawable/language_spinner_background.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle" >
+
+    <solid android:color="@color/black60alpha" />
+
+    <corners
+        android:bottomLeftRadius="24dp"
```
Cliabhach: The main difference in appearance here seems to be these 24dp corner radii. Have you checked how this new background looks with different text scaling?

sam33rdhakal: Yes, I tested it after your comments. Layout background is fine but I need to add some margin between the language text and caret sign. I will update it. Nich catch @Cliabhach. Thanks.

Cliabhach: Neat.

Cliabhach: ....if that solution will take too long we can merge it in as-is. Just need to add a TODO and rebase on `dev`.

sam33rdhakal: Updated.

Cliabhach: All right! I've tried it out locally and it looks ok there too.

PR #330, opened 2018-11-07
Fixes map tile image cache issue

sam33rdhakal:
* Fixes the tile image cache issue (AIC-610)
* Removes the artic objects before updating so that removed objects in server are also removed by app (AIC-611)

### DIFF
--- a/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
+++ b/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
@@ -147,11 +147,13 @@ class AppDataManager @Inject constructor(
 
                             val galleries = result.galleries?.values?.toList()
                             if (galleries?.isNotEmpty() == true) {
+                                galleryDao.clear()
                                 galleryDao.addGalleries(galleries)
                             }
 
                             val audioFiles = result.audioFiles
                             if (audioFiles?.isNotEmpty() == true) {
+                                audioFileDao.clear()
                                 audioFileDao.addAudioFiles(audioFiles.values.toList())
                             }
 
--- a/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
+++ b/db/src/main/kotlin/edu/artic/db/AppDataManager.kt
@@ -169,6 +169,7 @@ class AppDataManager @Inject constructor(
                                         }
                                     }
                                 }
+                                objectDao.clear()
                                 objectDao.addObjects(objects.values.toList())
                             }
 
--- a/image/build.gradle
+++ b/image/build.gradle
@@ -4,6 +4,6 @@ dependencies {
     kapt libs.glide_compiler
 
     api libs.rx_helpers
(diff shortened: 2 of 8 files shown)


image/build.gradle
```diff
@@ -4,4 +4,6 @@ dependencies {
     kapt libs.glide_compiler
 
     api libs.rx_helpers
+    implementation ('com.github.bumptech.glide:okhttp3-integration:4.4.0@aar')
```
Cliabhach: Why aren't you using the same `glide_version` defined in the top-level build.gradle?

image/src/main/java/edu/artic/image/GlideModule.kt
```diff
@@ -44,6 +54,25 @@ class GlideModule : AppGlideModule() {
...
+         * Configuration to use OkHttp client as the network client for Glide.
+         * OkHttp is also configured here to cache the image.
+         * Glide doesn't show the latest image if its from the same URL.
+         * OkHttp solves this problem by caching the image considering E-Tag and Last-Modified headers.
+         *
+         * @see https://github.com/bumptech/glide/issues/1847
+         */
+        val cacheSize: Long = 10 * 1024 * 1024 // 10 MiB
+        val clientBuilder = OkHttpClient.Builder()
+                .addInterceptor(HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.HEADERS))
```
Cliabhach: Please do not enable logging in release builds.

image/build.gradle
```diff
@@ -4,4 +4,6 @@ dependencies {
...
+    implementation 'com.squareup.okhttp3:logging-interceptor:3.6.0'
```
Cliabhach: Please use `debugImplementation` or similar.

db/src/main/kotlin/edu/artic/db/AppDataManager.kt
```diff
@@ -169,6 +169,7 @@ class AppDataManager @Inject constructor(
                                         }
                                     }
                                 }
+                                objectDao.clear()
```
Cliabhach: Based on this, do we need to clear out the gallery and audio file DAOs as well?

image/build.gradle
sam33rdhakal: Updated.

image/build.gradle
sam33rdhakal: Nice catch! Updated

image/src/main/java/edu/artic/image/GlideModule.kt
sam33rdhakal: Now interceptor will only be added if build is `Debug`.

image/build.gradle
Cliabhach: huh. forgot we couldn't use that without splitting out code. guess it's fine as-is

PR #334, opened 2018-11-08
Use new property for exhibition images (AIC-612)

sam33rdhakal:
* Adds the new attribute `image_url` to `ArticExhibition`
* Updates the view logic so that imageView's size is used to build the image URL 
`url=baseurl?w=imageView.measuredWidth&h=imageView.measuredHeight`

### DIFF
--- a/content_listing/src/main/kotlin/edu/artic/exhibitions/AllExhibitionsAdapter.kt
+++ b/content_listing/src/main/kotlin/edu/artic/exhibitions/AllExhibitionsAdapter.kt
@@ -8,6 +8,7 @@ import edu.artic.adapter.AutoHolderRecyclerViewAdapter
 import edu.artic.adapter.BaseViewHolder
 import edu.artic.content.listing.R
 import edu.artic.image.GlideApp
+import io.reactivex.android.schedulers.AndroidSchedulers
 import kotlinx.android.synthetic.main.cell_all_exhibitions_layout.view.*
 
 /**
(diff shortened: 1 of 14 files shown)


db/src/main/kotlin/edu/artic/db/models/ArticExhibition.kt
```diff
@@ -44,6 +44,11 @@ data class ArticExhibition(
             return aic_start_at.toCurrentTimeZone()
         }
 
+    val thumbUrl: String?
+        get() {
+            return "$imageUrl?w=200&h=150"
```
Cliabhach: If `imageUrl` is nullable, should we maybe return null or an empty string if `imageUrl` is null?

sam33rdhakal: Yea. How did i miss that. Will update it. Thanks.

sam33rdhakal: Updated.

content_listing/src/main/kotlin/edu/artic/exhibitions/AllExhibitionsAdapter.kt
```diff
@@ -17,13 +18,18 @@ import kotlinx.android.synthetic.main.cell_all_exhibitions_layout.view.*
...
 
     override fun View.onBindView(item: AllExhibitionsCellViewModel, position: Int) {
-        item.exhibitionImageUrl.subscribe {
-            GlideApp.with(context)
-                    .load(it)
-                    .error(R.drawable.placeholder_medium_square)
-                    .placeholder(R.color.placeholderBackground)
-                    .into(image)
-        }.disposedBy(item.viewDisposeBag)
+        item.exhibitionImageUrl
```
Cliabhach: Do we have to worry about `exhibitionImageUrl` being empty here?

sam33rdhakal: If its empty we'll display the placeholder image. So we don't need to worry.

PR #337, opened 2018-11-08
Map Crash fix : (AIC-616)

sam33rdhakal:
`requireContext()` was returning exception. Now uses nullable context and only shows `infoWindow` if the context is not null.


map/src/main/kotlin/edu/artic/map/MapFragment.kt
```diff
@@ -687,6 +691,7 @@ class MapFragment : BaseViewModelFragment<MapViewModel>() {
     override fun onDestroyView() {
         mapView.onDestroy()
         leaveTourDialog?.dismissAllowingStateLoss()
+        infoAdapter = null
```
Cliabhach: ah, but, um, the GoogleMap object can still use the object here...don't we need to call `mapView.setInfoWindowAdapter(null)`?

Cliabhach: I've done just that, in `MapViewModel::cleanup`. @caguilar187 just a head's-up: we will need you to help verify this code before long.

PR #342, opened 2018-11-14
Displays badge for reciprocal member accounts (AIC-614)

sam33rdhakal:


access_card/src/main/res/layout/layout_member_access_card.xml
```diff
@@ -34,14 +35,28 @@
...
+        android:lines="1"
+        app:layout_constraintEnd_toStartOf="@id/reciprocalMember"
         app:layout_constraintHorizontal_bias="0"
         app:layout_constraintStart_toStartOf="parent"
         app:layout_constraintTop_toBottomOf="@id/membershipType"
-        tools:text="Member Plus" />
+        tools:text="Member Plus Member Plus Member Plus Member Plus" />
+
+    <ImageView
+        android:id="@+id/reciprocalMember"
```
Cliabhach: Shouldn't the default visibility be `GONE`? How are we hiding it if the membership is not reciprocal?

access_card/src/main/java/edu/artic/accesscard/AccessMemberCardFragment.kt
```diff
@@ -167,6 +167,10 @@ class AccessMemberCardFragment : BaseViewModelFragment<AccessMemberCardViewModel
                 .map { it.size > 1 }
                 .bindToMain(switchCardHolder.visibility())
                 .disposedBy(disposeBag)
+
+        viewModel.isReciprocalMemberLevel
+                .bindToMain(reciprocalMember.visibility())
```
Cliabhach: What do you think of assigning the `memberInfo.memberLevel` as contentDescription?

access_card/src/main/res/layout/layout_member_access_card.xml
sam33rdhakal: Yes, it should be `GONE`. Updated.

access_card/src/main/java/edu/artic/accesscard/AccessMemberCardFragment.kt
sam33rdhakal: Done.

PR #343, opened 2018-11-14
Hides empty sections in audio player details : (AIC-617)

sam33rdhakal:
When there is no data for `related tours` & `artist culture place`, this PR hides those sections from `audio details screen`.

### DIFF
--- a/media_ui/src/main/java/edu/artic/media/ui/AudioDetailsFragment.kt
+++ b/media_ui/src/main/java/edu/artic/media/ui/AudioDetailsFragment.kt
@@ -133,7 +133,16 @@ class AudioDetailsFragment : BaseViewModelFragment<AudioDetailsViewModel>() {
 
         viewModel.authorCulturalPlace
                 .map { it.isNotEmpty() }
-                .bindToMain(artistCulturePlaceDenim.visibility())
+                .observeOn(AndroidSchedulers.mainThread())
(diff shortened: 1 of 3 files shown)


media_ui/src/main/java/edu/artic/media/ui/AudioDetailsFragment.kt
```diff
@@ -165,15 +170,12 @@ class AudioDetailsFragment : BaseViewModelFragment<AudioDetailsViewModel>() {
...
-                        relatedToursView.removeAllViews()
-                        relatedTourTitle.visibility = View.VISIBLE
-                        relatedToursView.visibility = View.VISIBLE
-                        dividerBelowRelatedTours.visibility = View.VISIBLE
+                    val hasData = tours.isNotEmpty()
+                    relatedTourTitle.show(show = hasData)
+                    relatedToursView.show(show = hasData)
+                    dividerBelowRelatedTours.show(show = hasData)
+
+                    if (hasData) {
```
Cliabhach: Just to clarify, we do not need to call `relatedToursView.removeAllViews()` anymore?

sam33rdhakal: My bad. That def should be there. Thank you for catching it @Cliabhach

PR #347, opened 2018-11-15
Focuses tour intro pin properly when the map is launched in tour mode - (AIC-613)

sam33rdhakal:
* Focuses tour intro pin properly when the map is launched in tour mode (AIC-613).
* Updates tour stop in carousel and map when the user selects new tour stop from tour details page without changing tour.


map/src/main/kotlin/edu/artic/map/MapViewModel.kt
```diff
@@ -661,6 +661,13 @@ class MapViewModel @Inject constructor(val mapMarkerConstructor: MapMarkerConstr
                         searchManager.selectedObject.onNext(Optional(null))
                         val startStop = requestedTourStop ?: activeTour.getIntroStop()
                         switchTourRequest.onNext(requestedTour to startStop)
+                    } else if (requestedTour != null && activeTour != null && requestedTour == activeTour) {
+                        /**
+                         * Update the tour stop.
+                         */
+                        requestedTourStop?.objectId?.let { newTourStop ->
```
Cliabhach: At some point we should consider making `objectId` here a non-null property.

PR #349, opened 2018-11-16
Displays proper error message for network errors

sam33rdhakal:
Updates
* Displays no internet connection for network error message
* Displays something went wrong when the error message is blank
* Updates exhibition search service to include new `image_url` attribute


db/src/main/kotlin/edu/artic/ResultExtensions.kt
```diff
@@ -16,19 +16,21 @@ import org.json.JSONObject
...
     lateinit var errorMessage: String
     try {
         this.response().errorBody()?.also { errorBody ->
             val errorJSON = JSONObject(errorBody.string())
-            errorMessage = errorJSON.optString("error")
+            val message = errorJSON.optString("error")
                     .orIfNullOrBlank(errorJSON.optString("message"))
                     .orIfNullOrBlank(errorJSON.optString("Message"))
                     .orIfNullOrBlank(errorJSON.optString("detail"))
-                    ?: "Something went wrong"
```
Cliabhach: What are your thoughts on using `.orIfNullOrBlank(SOME_THING_WENT_WRONG)` here?

db/src/main/kotlin/edu/artic/ResultExtensions.kt
```diff
@@ -16,19 +16,21 @@ import org.json.JSONObject
  * Looks for error, message and detail keys in order.
  */
 fun Result<*>.getErrorMessage(): String {
+    val SOME_THING_WENT_WRONG : String = "Something went wrong"
```
Cliabhach: `SOME_THING_WENT_WRONG` -> `SOMETHING_WENT_WRONG`

db/src/main/kotlin/edu/artic/ResultExtensions.kt
sam33rdhakal: `orIfNullOrBlank` returns `String?` and `errorMessage` is of type `String`.

sam33rdhakal: I can use bang operator here if i use `.orIfNullOrBlank(SOME_THING_WENT_WRONG)`
```
val message = errorJSON.optString("error")
                    .orIfNullOrBlank(errorJSON.optString("message"))
                    .orIfNullOrBlank(errorJSON.optString("Message"))
                    .orIfNullOrBlank(errorJSON.optString("detail"))
                    .orIfNullOrBlank(SOMETHING_WENT_WRONG)
            errorMessage = message!!
```

db/src/main/kotlin/edu/artic/ResultExtensions.kt
sam33rdhakal: Updated.

db/src/main/kotlin/edu/artic/ResultExtensions.kt
sam33rdhakal: Ok.

PR #350, opened 2018-11-20
Submits sign in form upon pressing enter in keyboard - (AIC-619)

sam33rdhakal:
Updates:
* Adds next action key to memberID which takes the user to the zipCode field
* Adds done action key to zipCode which submits the form  (only if form is filled)


access_card/src/main/java/edu/artic/accesscard/AccessMemberCardFragment.kt
```diff
@@ -226,6 +227,18 @@ class AccessMemberCardFragment : BaseViewModelFragment<AccessMemberCardViewModel
...
+        zipCode.setOnEditorActionListener { _, actionId, _ ->
+            when (actionId) {
+                EditorInfo.IME_ACTION_DONE -> {
+                    val isFormFilled = !zipCode.text.isNullOrBlank() && !memberId.text.isNullOrBlank()
+                    if (isFormFilled) {
+                        viewModel.onSignInClick()
+                    }
+                    true
+                }
+                else -> false
```
Cliabhach: I suppose in the future we should consider adding an error message or something to highlight zip and/or member id if they're empty.

access_card/src/main/res/layout/layout_member_information_form.xml
```diff
@@ -63,6 +65,9 @@
         android:autofillHints="postalCode"
         android:background="@color/infoInputFieldWhite"
         android:padding="@dimen/marginStandard"
+        android:imeOptions="actionDone"
+        android:inputType="textPostalAddress"
```
Cliabhach: Good idea! This will support non-US addresses.

access_card/src/main/java/edu/artic/accesscard/AccessMemberCardFragment.kt
sam33rdhakal: Currently, sign in button is disabled if the form is not filled.

PR #352, opened 2018-11-20
Search: Fixes the spacing issue between suggested artworks (AIC-633)

sam33rdhakal:
The cause behind this issue was the logic we used to get column number for suggested artworks. That logic was based on the position of the `on the map` cell.

```
val columnValue = (currentPosition - onTheMapPosition) % maxSize
```

This logic can go wrong when suggested artwork cells are recycled before that `on the map` cell. So the proposed solution is to not use `on the map` cell at all. Now I pass the order of the cells when they are constructed and then use it to calculate the column index. 

```
val columnValue = viewModel.order % maxSize
```

### DIFF
--- a/search/src/main/java/edu/artic/search/SearchResultCellViewModels.kt
+++ b/search/src/main/java/edu/artic/search/SearchResultCellViewModels.kt
@@ -86,7 +86,7 @@ class SearchTourCellViewModel(val articTour: ArticTour) : SearchBaseListItemView
  * ViewModel for displaying the circular artwork image under "On the map" section.
  */
 class SearchCircularCellViewModel(val artwork: ArticObject?,
-                                  override val order: Int = -1) : SearchBaseCellViewModel(), OrderedCellViewModel {
+                                  override val order: Int) : SearchBaseCellViewModel(), OrderedCellViewModel {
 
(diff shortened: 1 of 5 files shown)


search/src/main/java/edu/artic/search/SearchResultCellViewModels.kt
```diff
@@ -85,7 +85,8 @@ class SearchTourCellViewModel(val articTour: ArticTour) : SearchBaseListItemView
 /**
  * ViewModel for displaying the circular artwork image under "On the map" section.
  */
-class SearchCircularCellViewModel(val artwork: ArticObject?) : SearchBaseCellViewModel() {
+class SearchCircularCellViewModel(val artwork: ArticObject?,
+                                  override val order: Int = -1) : SearchBaseCellViewModel(), OrderedCellViewModel {
```
Cliabhach: I think we can remove the default value for `order` here. That way the compiler will error out if we forget to include a number.

Cliabhach (changes requested):
One comment; everything else looks great.

Cliabhach (approved):
Great! Let's get this in.

nikhiltri (approved):
Good fix. ⭐

PR #353, opened 2018-11-20
Update gradle dependencies for November 2018

Cliabhach:
This is the second in our monthly series of regular maintenance changesets (you can find the October 2018 one under PR #311). The November changes deal primarily with RX dependencies.

While the functionality of output APKs should not be different, the build speed should be a little faster and the dependency graph a little easier to read.

Included in this PR:

# Introduce new `:reactivex` module
# Remove `android.arch.persistence.room:rxjava2` from all but the `:db` module

Planned (but not yet included):

* Reduce complexity of `:map` module dependency graph
* Restrict `:base` dependencies to the absolute minimum
* Suppress that annoying configure-time warning (`Some runtime JAR files in the classpath have an incompatible version. Consider removing them from the classpath`)


Cliabhach:
...Although, on second thought, that might take a little while longer to get right. Moving the logic to `:media_ui` can be a separate PR. I'll update the description here accordingly.

Cliabhach:
I'm going to drop the changes to `:base` from this PR. That can be covered in December or January instead.

sam33rdhakal:
@Cliabhach Feel free to merge it.

PR #355, opened 2019-01-29
Fixes issues related to audio lookup

sam33rdhakal:
Fixes 
* AIC-640: Some audio guide numbers not working on Android
* AIC-641: Tour selector numbers not working in Android

### DIFF
--- a/audio/src/main/java/edu/artic/audio/AudioLookupViewModel.kt
+++ b/audio/src/main/java/edu/artic/audio/AudioLookupViewModel.kt
@@ -43,6 +43,7 @@ import javax.inject.Inject
 class AudioLookupViewModel @Inject constructor(
         private val analyticsTracker: AnalyticsTracker,
         objectLookupDao: ArticObjectDao,
+        tourLookupDao: ArticTourDao,
         generalInfoDao: GeneralInfoDao,
         languageSelector: LanguageSelector,
         private val audioPrefManager: AudioPrefManager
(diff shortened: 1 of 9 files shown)


audio/src/main/java/edu/artic/audio/AudioLookupViewModel.kt
```diff
@@ -106,20 +111,29 @@ class AudioLookupViewModel @Inject constructor(
 
         val objectFoundObservable = lookupRequests
                 .observeOn(Schedulers.io())
-                .map {
-                    Optional(objectLookupDao.getObjectBySelectorNumber(it))
+                .map { objectSelectorNumber ->
+
+                    val likeCondition = """
+                        %"object_selector_number":"$objectSelectorNumber"%
+                    """.trimIndent()
```
Cliabhach: I'm not entirely happy with having this condition here. Wouldn't it make more sense for that to be a constant in the DAO interface or the model class?

media_ui/src/main/java/edu/artic/media/ui/AudioDetailsViewModel.kt
```diff
@@ -46,7 +47,9 @@ class AudioDetailsViewModel @Inject constructor(
     val authorCulturalPlace: Subject<String> = BehaviorSubject.create()
     private val objectObservable: Subject<Playable> = BehaviorSubject.create()
     val relatedTours: Subject<List<ArticTour>> = BehaviorSubject.create()
-
+    val currentAudioFileModel: Subject<ArticAudioFile> = BehaviorSubject.create()
```
Cliabhach: `currentAudioFileModel` -> `currentAudioFile`

media/src/main/java/edu/artic/media/audio/AudioPlayerService.kt
```diff
@@ -502,6 +502,26 @@ class AudioPlayerService : DaggerService(), PlayerService {
         analyticsTracker.reportEvent(EventCategoryName.PlayBack, AnalyticsAction.playbackInterrupted, (currentTrack as BehaviorSubject).value?.title.orEmpty())
         audioControl.onNext(AudioPlayerService.PlayBackAction.Stop())
     }
+
+    /**
+     * Returns currently playing [ArticAudioFile].
```
Cliabhach: Hmm. May be worth mentioning in the comment that tours don't need/have these.

audio/src/main/java/edu/artic/audio/AudioLookupViewModel.kt
sam33rdhakal: Good idea @Cliabhach. Moved it to DAO interface.

media_ui/src/main/java/edu/artic/media/ui/AudioDetailsViewModel.kt
sam33rdhakal: Refactored.

media/src/main/java/edu/artic/media/audio/AudioPlayerService.kt
sam33rdhakal: Added the docs mentioning that method is intended to be used for `ArticObject` only.

PR #356, opened 2019-02-05
Disable the transition animation in lollipop devices

sam33rdhakal:
Shared transition element does not work properly in Lollipop and crashes on some devices. So, the transition animation is disabled for those devices.

### DIFF
--- a/splash/src/main/kotlin/edu/artic/splash/SplashActivity.kt
+++ b/splash/src/main/kotlin/edu/artic/splash/SplashActivity.kt
@@ -9,6 +9,7 @@ import android.graphics.Matrix
 import android.graphics.SurfaceTexture
 import android.media.MediaPlayer
 import android.os.Build
+import android.os.Build.VERSION_CODES
 import android.os.Bundle
 import android.support.annotation.UiThread
 import android.transition.Fade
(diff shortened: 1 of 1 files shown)


splash/src/main/kotlin/edu/artic/splash/SplashActivity.kt
```diff
@@ -225,8 +226,18 @@ class SplashActivity : BaseViewModelActivity<SplashViewModel>(), TextureView.Sur
...
                 .makeSceneTransitionAnimation(this, textureView, "museumImage")
         if (!isFinishing) {
-            startActivity(intent, options.toBundle())
-            finishAfterTransition()
+            /**
+             * Shared transition element does not work properly (crashes on some devices) running
+             * Lollipop. Please check https://issuetracker.google.com/issues/35826109
+             */
+            if (Build.VERSION.SDK_INT < VERSION_CODES.M &&
+                    Build.VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP) {
```
Cliabhach: I think this is all acceptable. Although...why not just check against Marshmallow?

sam33rdhakal: Agreed. Removed the `Build.VERSION.SDK_INT >= VERSION_CODES.LOLLIPOP` condition since min support is `VERSION_CODES.LOLLIPOP`

PR #358, opened 2019-02-11
Manually hide transcript & credit views on Tours' Audio Details screen (AIC-646)

sam33rdhakal:
This changes the logic slightly: the `transcript` and `credit` ViewGroups are now visible by default. This means, of course, that we need to hide them if the current `Playable` is a Tour....


sam33rdhakal:
I am working to figure out the root cause 0-height measurement issue. Height of `DropDownView` is not measured correctly only when it's initial visibility in audio_details layout is `View.GONE`. 
This issue does not happen when the initial visibility is `View.VISIBLE`.

PR #361, opened 2019-02-13
Remove unused mandatory fields : AIC-648

sam33rdhakal:
Removes unused non-optional fields from the model classes.

### DIFF
--- a/db/src/main/kotlin/edu/artic/db/models/ArticObject.kt
+++ b/db/src/main/kotlin/edu/artic/db/models/ArticObject.kt
@@ -31,8 +31,6 @@ data class ArticObject(
         @Json(name = "artwork_type_id") val artworkTypeId: String? = "",
         @Json(name = "credit_line") val creditLine: String? = "",
         @Json(name = "copyright_notice") val copyrightNotice: String? = "",
-        @Deprecated("Unused")
-        @Json(name = "in_gallery") val inGallery: Boolean = false,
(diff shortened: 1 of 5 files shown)


db/src/main/kotlin/edu/artic/db/models/ArticObject.kt
```diff
@@ -52,12 +51,9 @@ data class ArticObject(
         @Json(name = "reference_num") val referenceNum: String? = null,
         @Json(name = "audio_commentary") val audioCommentary: List<AudioCommentaryObject> = listOf(),
         @Json(name = "highlighted_object") val highlightedObject: String? = null,
-        @Deprecated("Please do not use the 'full' image, as it is too large to fit on screen. Migrate to 'image_url' or 'large_image_full_path' immediately.")
-        @Json(name = "full_image_full_path") val fullImageFullPath: String? = null,
 //        @Json(name = "audio") val audio: List<String?>, Removed for now until json gets fixed as sometimes returns string othertimes object
```
Cliabhach: This might be a good time to remove this too? We're retrieving audio properties in a different way (`audioCommentary`) now.

sam33rdhakal: @Cliabhach Removed it.

sam33rdhakal:
Totally agree @Cliabhach, let do inline doc for the fields where the name itself is not sufficient.

PR #363, opened 2019-05-26
Update gradle dependencies for May 2019

Cliabhach:
This the third of our 'monthly' series of regular maintenance changesets (you can find the November 2018 one under #353 and the October 2018 one under PR #311). The May changes focus on upgrading the most outdated dependencies.

Of special note is that we now use [my artifact](https://github.com/Cliabhach/okhttp3-downloadprogress-interceptor/releases/tag/1.0.2-a1) for the Okhttp3 Progress Interceptor dependency.

Included in this PR:
1. Gradle release 5.4.1 (previously 4.10.1)
2. Kotlin 1.3.31 (previously 1.2.71)
3. Kotlin Annotation Processor (`KAPT`) will now run as we intended on JDK 9+ hosts
4. Explicit dependency on Kotlin's reflection library
5. Limited Dagger2 update from 2.15 to 2.19
6. Disable two annoying lint checks

We really will need to address the `ConstraintLayout` lint checks sooner rather than later. That and the `:media_ui` changes referenced in the November 2018 gradle updates should be a priority for June 2019.


Cliabhach:
As came up in offline discussion, we'll need to address the ContraintLayout changes sooner rather than later. I'll open a new ticket for that.

PR #218, opened 2018-10-05
Improvements to the overall audio experience when using the notification

caguilar187:

PR #366, opened 2019-05-28
Permit system control of audio (AIC-651)

Cliabhach:
Before, we restricted playback to the earpiece at exclusion of all other connections (c.f. `AIC-271`, `AIC-272`, #218 ). It turned out that didn't work so well when headphones were plugged in, so we're disabling the behavior.

### DIFF
--- a/media/src/main/java/edu/artic/media/audio/AudioPlayerService.kt
+++ b/media/src/main/java/edu/artic/media/audio/AudioPlayerService.kt
@@ -204,7 +204,7 @@ class AudioPlayerService : DaggerService(), PlayerService {
      *
      * @see AudioManager.MODE_IN_COMMUNICATION
      */
-    internal var forceHeadsetMode: Boolean = true
+    internal var forceHeadsetMode: Boolean = false
 
 
     override fun onCreate() {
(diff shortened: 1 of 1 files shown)


media/src/main/java/edu/artic/media/audio/AudioPlayerService.kt
```diff
@@ -185,6 +186,20 @@ class AudioPlayerService : DaggerService(), PlayerService {
...
+     * On Android Pie devices, that would let the system load a reasonable
+     * volume-settings preset. Only takes effect when a [Playing] event passes
+     * through [audioPlayBackStatus].
+     *
+     * Assumption: [AudioManager.setMode] only affects the mode for this service. When
+     * we die, the mode will revert to its previous setting.
+     *
+     * @see AudioManager.MODE_IN_COMMUNICATION
+     */
+    internal var forceHeadsetMode: Boolean = false
```
caguilar187: What's controlling this? Also wouldn't false break it for the more common use case?

Cliabhach: This isn't being controlled by anything right now - we could make it a val, or inline it.

Cliabhach: As discussed offline, changes to this PR to resolve the question of `forceHeadsetMode` are out of scope. We can revisit the question later.

PR #367, opened 2019-05-29
Avoid parsing null language tags (AIC-652)

Cliabhach:
The implementation of `SpecifiesLanguage::underlyingLocale()` will throw a NullPointerException if the `SpecifiesLanguage` object in question doesn't have a language. This PR simply returns the `Locale.ROOT`constant from that method in such a scenario, but realistically we shouldn't even allow the creation of objects without a defined language.

I've been working on a more holistic solution to this problem in a separate branch - `pc-better-active-audio-file-guarantees` - but I haven't scoped out the implications of showing a language selector on the tour intro page yet. Regardless, this PR should be fully capable of working around the crashes.

@caguilar187 I'd like you to review. I think we need to discuss whether there might be an alternative to the current use of `EMPTY_AUDIO_FILE` in `AudioPlayerService.kt`, as I believe that is the specific cause of the crash described in AIC-652.

### DIFF
--- a/localization/src/main/kotlin/edu/artic/localization/SpecifiesLanguage.kt
+++ b/localization/src/main/kotlin/edu/artic/localization/SpecifiesLanguage.kt
@@ -18,7 +18,17 @@ interface SpecifiesLanguage {
     fun underlyingLanguage() : String?
 
     fun underlyingLocale(): Locale {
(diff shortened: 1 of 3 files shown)


Cliabhach:
Cross-ref: #355

PR #375, opened 2020-02-08
Add rental product flavor

nuudles:
This pull request contains the commits necessary to add product flavors to the app so that it can be compiled for a "Rental" version intended for devices to be rented at the museum and a "Store" version intended for submission to the Google Play Store. It also contains some fixes for some issues discovered through an UI audit.

Please note that with these changes, `google-services.json` files for the Store version and the Rental version should be saved in the `app/src/store` and the `app/src/rental` directories respectively.


surreal8:
thanks @nuudles!

PR #381, opened 2025-06-17
Update map tiler script [MA-171]

zachgarwood:
- Clarify documentation
- Update commands to use most recent version of ImageMagick
- Update stdout messages
- Add check for required dependencies


nikhiltri (approved):
Looks good. ⭐

PR #384, opened 2025-06-29
Fix: Adjust welcome screen layout

khealy-dev:
This PR removes a TextView that was originally being populated by the `home_member_prompt_text` field from the CMS which is no longer being set, which caused the layout to look like it had too much top padding. We also increase the margins of the remaining TextView so it doesn't look awkwardly small on its own.

Before:

After:


nikhiltri (approved):
Nice. ⭐

PR #385, opened 2025-06-29
Fix: Add localized strings for museum hours

khealy-dev:
This PR adds localized strings for the museum hours field in the Museum Information screen for all languages. But since the CMS field is still populated for at least some languages, we'll continue to use that via the view binding if and only if it's not a blank string.

Before (e.g. Spanish):

After:


nikhiltri (dismissed):
Nice. ⭐ Where did you get the translated content from?

PR #386, opened 2025-06-29
Fix: Adjust attributes for autosizing numpad keys

khealy-dev:
This small PR just makes a couple of tweaks to the attributes of the numpad key's TextView, which was already set up to autosize its text, so that the French "Lancer" button won't be truncated to 2 characters or clip out of the key's bounds.

Before:

After:


nikhiltri (dismissed):
Nice! ⭐

nikhiltri (approved):

PR #389, opened 2025-07-03
Remove old Room database migration workaround 

khealy-dev:
One of the outdated dependencies that was updated [in this PR](https://github.com/art-institute-of-chicago/aic-mobile-android/pull/382/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R14) was Room, the database framework that our app uses to persist the CMS content to disk and populate the app. As it turns out, that update turned an old migration-enabling workaround into a migration-stopping bug! Fortunately, the fix was as simple as doing exactly what the comment said to do and removing the workaround. Thanks, past developer. 👍


nikhiltri (approved):
Nice! ⭐

PR #390, opened 2025-07-27
Feature: Remove suggested artworks from Search

khealy-dev:
This PR removes the "suggested artworks" functionality from search and search results, since it isn't implemented in the iOS app and it's not obvious to users what they're even looking at. It also removes some buggy code that attempted to manually adjust the padding of the item types that can be rendered side-by-side in the same row (now limited to just amenities), since it often caused them to overlay each other as the search results change and no longer seems to be necessary (maybe it never was? 🤔)

Before:
Screenshot_1753649850
Screenshot_1753649853

Example of buggy padding code:
Screenshot_1753650389

After:
Screenshot_1753649793
Screenshot_1753649799


nikhiltri (approved):
Looks good! ⭐

PR #391, opened 2025-08-05
Refactor `map/tiler.sh` script [MA-173]

zachgarwood:
- add another zoom level
- create variables for repeated strings
- move transient files to a directory in `/tmp`


nikhiltri (approved):
Nice updated output! ⭐ 

```
floor: 2 | zoom level: 2 | tiles: 16
floor: 2 | zoom level: 3 | tiles: 64
floor: 2 | zoom level: 4 | tiles: 256
generating tiles for zoom level 5...
```

PR #393, opened 2025-08-07
Feature: Update map zoom

khealy-dev:
This PR increases the maximum zoom of our map, allowing it to fetch the new zoom level 6 tiles. Note that Google Maps SDK technically allows a max zoom of 22 (even though Google won't let us use it today 🫠 ), while we set our max to just _under_ 22. This is because if we actually could reach 22 then the app would request zoom level 7 tiles, which aren't available. I also slightly adjusted the bounds of the tiles at zoom level 6 to better line them up with the real latitude and longitude values of the rooms and objects in the museum.

There are also two bug fixes for map-related crashes that are reported in the current production app:

1.  We use a third party library called [CircleImageView](https://github.com/hdodenhof/CircleImageView) to render our map markers. There was sometimes a crash inside that library when we tried to set the circle's border width but the image bitmap used in it had already been recycled. First, the version of the library we were using was years outdated and the code that crashed is no longer present in the latest version. But second, I've reordered the code in the `makeIcon` function so that a new bitmap is always set _before_ we make any other adjustments to the imageview.
2. In our TileProvider implementation, we were attempting to catch only `IOException`s when fetching new tiles from the server with Glide, our image fetching/caching library. If the exception is caught then we log it and return `null`, which is the expected value when a tile can't be fetched and signals Google Maps to try again with an exponential backoff. But the code inside the try block can't actually throw an `IOException`; instead it can throw `ExecutionException`s wrapping `GlideException`s or `InterruptedException`s, which are not caught and just crash the app. At any rate, the correct way to handle _any_ exception in the tile provider is to return null, so that's what we'll do.

### DIFF
--- a/map/src/main/kotlin/edu/artic/map/rendering/GlideMapTileProvider.kt
+++ b/map/src/main/kotlin/edu/artic/map/rendering/GlideMapTileProvider.kt
@@ -1,12 +1,12 @@
 package edu.artic.map.rendering
 
 import android.content.Context
+import android.util.Log
 import com.bumptech.glide.load.engine.DiskCacheStrategy
 import com.google.android.gms.maps.model.Tile
 import edu.artic.db.models.ArticMapFloor
 import edu.artic.image.GlideApp
 import timber.log.Timber
-import java.io.IOException
 import kotlin.math.pow
 
 /**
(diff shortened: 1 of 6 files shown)


map/src/main/kotlin/edu/artic/map/rendering/GlideMapTileProvider.kt
```diff
@@ -48,8 +51,8 @@ class GlideMapTileProvider(private val context: Context,
                             TILE_SIZE.toInt(),
                             array
                     )
-                } catch (e: IOException) {
-                    Timber.e("Could not find tile with $path")
+                } catch (e: Exception) {
+                    Timber.tag(TAG).log(Log.ERROR, e, "Could not find tile with $path")
```
zachgarwood, line 55: 👍

nikhiltri (approved):
Looks good! ⭐

PR #394, opened 2025-08-22
Feature: Merge main into dev

khealy-dev:
Just housekeeping to get everything back in order.


nikhiltri:
Hey @khealy-dev. I'm not seeing these two commits in this PR. Are these already in `main`?

Screenshot 2025-08-29 at 10 23 43 AM

khealy-dev:
> Hey @khealy-dev. I'm not seeing these two commits in this PR. Are these already in `main`?
> 
> Screenshot 2025-08-29 at 10 23 43 AM

@nikhiltri My thinking was that since Zach's changes weren't in the release branch that I built the current store build off, it would make more sense to leave those commits just in develop until I make the next release. From my POV as a mobile app guy I personally prefer for `main` to contain just what's in the store build, but that is admittedly a little blurry here because the tiler script isn't part of the app and Zach's changes _were_ used to generate the tiles that the store build downloads. So really, I think I could go either way here; your call!

PR #396, opened 2026-03-23
Feature: Show boosted off view search results

khealy-dev:
This PR makes several updates to our Search Result data models and business logic so that we can show artwork which is off-view but boosted, and so that we have an accurate latitude and longitude for showing search results on the map if they _are_ on-view but weren't already in the app data.


nikhiltri (dismissed):
Nice. ⭐

PR #397, opened 2026-04-17
Feature: Update the app icon

khealy-dev:
This is a straightforward PR to update the app launcher icon across various sizes and settings. All of these images and drawables were generated with Android Studio's built-in tool [Image Asset Studio](https://developer.android.com/studio/write/create-app-icons#about) using the graciously provided .SVG layers. Note that the `ic_launcher-playstore.png` file isn't actually included when we build the app, but it makes sense to check it into source control since it was generated with the exact same settings as the launcher icons.


nikhiltri (approved):
Nice. ⭐

PR #398, opened 2026-04-17
Fix: Opt out of edge-to-edge enforcement on search & other screens

khealy-dev:
In our previous release, I added flags to our most commonly used Themes to opt us out of edge-to-edge enforcement on the newest versions of Android (essentially the expectation that our UI gracefully covers the entire device screen, accounting for camera cutouts, navigation and status bars, etc). However, I missed that some screens declared their own Themes as well, leading to some ugly clipping and whiteboxing on the search page among others. This PR ensures that all Themes used in our app uniformly opt out of edge-to-edge enforcement.

Before:
Screenshot_1776402220

After:
Screenshot_1776402256

Note, however, that this is not a long-term solution. Opting out won't be allowed once we're required to increase our target SDK to 36, which will happen at the end of August, so we'll need to invest effort in truly updating the UI before then.

### DIFF
--- a/audio/src/main/res/values-v35/styles.xml
+++ b/audio/src/main/res/values-v35/styles.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="AppTheme.CustomToolbar.AudioTheme">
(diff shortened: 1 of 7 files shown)


nikhiltri (approved):
Looks good. ⭐

PR #399, opened 2026-04-23
Feature: Improve the app startup experience

khealy-dev:
This PR groups a couple of related tasks together. First, we want to only show the startup video once after the app is installed; future launches should navigation to the welcome screen as soon as necessary data is downloaded. While implementing that and fixing the transition animation to account for the video no longer being shown, I noticed our navigation tab transitions were incorrectly animating (we want the tabs to swap instantly) on newer Android versions because of an SDK change, so I addressed that by overriding the animations using the new API. Finally, there was an old bug where sometimes when the app launches its lifecycle callbacks will be called before the system considers it fully in the foreground and it tries to start our audio service, which crashes if called from the background. This is just a Weird Android Thing, so we use the approach recommended by the Android team and check our process info before starting said service.

Before:
https://github.com/user-attachments/assets/cf10d3db-17ad-42ea-b2f7-e473144bb4c2

After:
https://github.com/user-attachments/assets/2670c1a8-4340-4b87-b6df-ec511fba4cb9

### DIFF
--- a/audio/src/main/java/edu/artic/audio/AudioActivity.kt
+++ b/audio/src/main/java/edu/artic/audio/AudioActivity.kt
@@ -7,6 +7,7 @@ import edu.artic.base.utils.preventReselection
 import edu.artic.media.ui.NarrowAudioPlayerFragment
 import edu.artic.navigation.NavigationSelectListener
 import edu.artic.navigation.linkHome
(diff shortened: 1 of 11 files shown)


nikhiltri (approved):
Looks good! ⭐

PR #401, opened 2026-05-11
Feature: Update Exhibitions to match other platforms

khealy-dev:
This PR updates the query that Android uses to fetch Exhibitions as well as removing the outdated client-side sorting logic referencing the CMS order; we now sort Exhibitions according to their `position` property from the API, making the Android experience totally consistent with iOS and much closer to the website's. The query that Android was previously using was filtering out any Exhibitions without an end date, so we also had to update the UI logic a little bit to allow for null end dates and show "Ongoing" when the end date is null; this is also consistent with iOS.

Before:

https://github.com/user-attachments/assets/bc90115b-d1b3-40aa-abf3-faae85489c74

After:

https://github.com/user-attachments/assets/8619eb34-835d-4625-9615-a96b864ce842

### DIFF
--- a/base/src/main/res/values-en/strings.xml
+++ b/base/src/main/res/values-en/strings.xml
@@ -51,4 +51,5 @@
 	<string name="messages_next_action">Next Message</string>
 	<string name="messages_previous_action">Previous Message</string>
 	<string name="search_currently_off_view">Currently Off View</string>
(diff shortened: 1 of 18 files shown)


nikhiltri (approved):
Looks good. ⭐

PR #402, opened 2026-05-11
Fix: Address app bar layout issues

khealy-dev:
There was an existing production bug in the Android app which seemingly isn't reproducible on all devices/versions that sometimes caused padding to be inserted at the top of the nested scrollview that all of our content renders in on the Welcome and Info pages. After some digging, it looks like this was caused by an internal bug in the [app bar's scrolling view behavior](https://developer.android.com/reference/com/google/android/material/appbar/AppBarLayout.ScrollingViewBehavior). Updating the Material Design dependency fixed that, but it broke an old workaround for customizing the Bottom Navigation Bar. Fortunately, the updated dependency now has direct APIs for that and we no longer need the workaround at all! 

Before:
Screenshot_20260510_235031

After:
Screenshot_20260510_235207

As a bonus, while digging into the app bar layout I also noticed that the Map screen's search icon wasn't properly aligned with the search icons on other screens because it's not actually part of the app bar, it's just a standalone element in the layout and it wasn't respecting the system insets. That's now fixed as well.

Before:
Screenshot_20260510_235754

After:
Screenshot_20260510_235535

### DIFF
--- a/audio/src/main/java/edu/artic/audio/AudioActivity.kt
+++ b/audio/src/main/java/edu/artic/audio/AudioActivity.kt
@@ -2,7 +2,6 @@ package edu.artic.audio
 
 import android.os.Bundle
 import edu.artic.audio.databinding.ActivityAudioBinding
-import edu.artic.base.utils.disableShiftMode
 import edu.artic.base.utils.preventReselection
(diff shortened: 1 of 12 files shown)


nikhiltri (approved):
Looks good. ⭐

